PR #3, opened 2024-06-04
Fix typo for Database creation

ncoop57:


ncoop57:
ah whoops sorry need to modify this in the nbs, not py

jph00:
No it's meant to be lower case -- that's a shortcut in `fastlite` that also creates the parent directory.

PS: please at-mention me in PRs etc so I get notified -- I don't want to miss your stuff :D

PR #9, opened 2024-06-09
Live Reloading

comhar:
This PR includes a basic implementation of live reloading.

How does the implementation work?
A subclass of FastHTML (FastHTMLWithLiveReload) does the following: 
  - adds a websocket at `/live-reload`
  - injects a small javascript snippet `LIVE_RELOAD_SCRIPT` to each webpage
  - this snippet allows the browser and server to communicate with each other
  - as a result the browser can "detect" code changes on the server and reload

Here are some items to resolve:
- [x] more robust reloading
- [x] unit tests
- [x] documentation
- [x]  test reloading with local static file

### DIFF
--- a/fasthtml/live_reload.py
+++ b/fasthtml/live_reload.py
@@ -22,6 +22,7 @@ LIVE_RELOAD_SCRIPT = """
     })();
 """
 
+
 async def live_reload_websocket(websocket):
     await websocket.accept()
 
@@ -54,6 +55,7 @@ class FastHTMLWithLiveReload(FastHTML):
     LIVE_RELOAD_ROUTE = WebSocketRoute("/live-reload", endpoint=live_reload_websocket)
 
     def __init__(self, *args, **kwargs):
-        kwargs.setdefault("hdrs", []).append(self.LIVE_RELOAD_HEADER)
-        kwargs.setdefault("routes", []).append(self.LIVE_RELOAD_ROUTE)
+        # "hdrs" and "routes" can be missing, None, a list or a tuple.
+        kwargs["hdrs"] = [*(kwargs.get("hdrs") or []), self.LIVE_RELOAD_HEADER]
+        kwargs["routes"] = [*(kwargs.get("routes") or []), self.LIVE_RELOAD_ROUTE]
         super().__init__(*args, **kwargs)
--- a/fasthtml/live_reload.py
+++ b/fasthtml/live_reload.py
@@ -3,17 +3,24 @@ from fasthtml import FastHTML, Script
 
 __all__ = ["FastHTMLWithLiveReload"]
 
-# TODO:
-#  making a single attempt to reload the browser 1 second after the `onclose` event
-#  is pretty basic. Let's look at some existing implementations and see if we can
-#  make the reload more robust.
+
 LIVE_RELOAD_SCRIPT = """
     (function() {
         var socket = new WebSocket(`ws://${window.location.host}/live-reload`);
-        socket.onclose = function() {setTimeout(window.location.reload(), 1000)};
+        var maxReloadAttempts = 20;
+        var reloadInterval = 250; // time between reload attempts in ms
+        socket.onclose = function() {
+            let reloadAttempts = 0;
+            const intervalFn = setInterval(function(){
+                window.location.reload();
+                reloadCount++;
+                if (reloadAttempts === maxReloadAttempts) {
+                    clearInterval(intervalFn);
+                };
+            }, reloadInterval);
+        }
     })();
 """
-
 
 async def live_reload_websocket(websocket):
     await websocket.accept()
@@ -29,15 +36,12 @@ class FastHTMLWithLiveReload(FastHTML):
       - a websocket is creaetd at `/live-reload`
       - a small js snippet `LIVE_RELOAD_SCRIPT` is injected into each webpage
       - this snippet connects to the websocket at `/live-reload` and listens for an `onclose` event
-      - when the onclose event is detected the browser reloads after 1 second
+      - when the onclose event is detected the browser is reloaded
 
     Why do we listen for an `onclose` event?
       When code changes are saved the server automatically reloads if the --reload flag is set.
       The server reload kills the websocket connection. The `onclose` event serves as a proxy
       for "developer has saved some changes".
-
-    Why do we wait for 1 second before reloading the browser?
-      We need to allow some time for uvicorn to reload and start serving the latest changes.
 
     Usage
         >>> from fasthtml.all import *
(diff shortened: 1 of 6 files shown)


comhar:
@jph00

settings.ini
```diff
@@ -4,7 +4,7 @@ lib_name = fasthtml
 version = 0.0.11
 min_python = 3.10
 license = apache2
-requirements = fastcore>=1.5.45 python-dateutil starlette oauthlib itsdangerous uvicorn httpx fastlite>=0.0.6 python-multipart
+requirements = fastcore>=1.5.45 python-dateutil starlette oauthlib itsdangerous uvicorn[standard] httpx fastlite>=0.0.6 python-multipart sqlite-utils
```
comhar, line 7: I was getting a missing package error for `sqlite_utils` so I added `sqlite-utils`.

jph00, line 7:
```suggestion
requirements = fastcore>=1.5.45 python-dateutil starlette oauthlib itsdangerous uvicorn[standard] httpx fastlite>=0.0.6 python-multipart
```

It shouldn't be necessary - I'm guessing one of your deps might need updating? We've switched to using sqlite-minutils now, which is a dep on fastlite.

jph00:
Looks like a nice approach! :) I can fix the routes TODO after merging, so don't worry about that for now.

settings.ini
comhar: discussed and resolved [here](https://github.com/AnswerDotAI/fasthtml/pull/9#discussion_r1632399002)

comhar:
## reload logic
I've updated the reload logic so that it attempts a reload every 250ms up to 5 seconds before giving up. Hopefully that covers most use cases 🤞 

## local static reloads
I tested `FastHTMLWithLiveReload` with some local static files (js, css).  To my pleasant surprise it handles this use-case as well 😄  

Here's the setup I used for those tests:

- create a static folder in the same directory as your app's main script.
static/
    - example.js
    - example.css

- use the following script
<details>
<summary>app script</summary>

```
from fasthtml.all import *
import uvicorn

STATIC_DIR = Path(__file__).parent / "static"
routes = [
    Mount('/static', app=StaticFiles(directory=STATIC_DIR), name="static"),
]

hdrs = [
    Script(src="static/logger.js"),
    Link(rel="stylesheet", type="text/css", href="static/main.css"),
]
app = FastHTMLWithLiveReload(hdrs=hdrs, routes=routes)
rt = app.route


@rt("/")
def get():
    return Title("FastHTML"), H1("Live reloading...")


if __name__ == '__main__':
    uvicorn.run(app, host='0.0.0.0', port=int(os.getenv("PORT", default=8000)))
```
⚠️ The same routes AttributeError that was thrown for WebSocketRoute is also thrown for `Mount`
</details>

- run the following uvicorn command
`uvicorn main:app --reload --reload-dir=../fasthtml --reload-include="*.py" --reload-include="*.js" --reload-include="*.css"`


## tests & documentation
thoughts on what tests and docs should be included?

fasthtml/live_reload.py
```diff
@@ -0,0 +1,55 @@
...
+        >>> app = FastHTMLWithLiveReload()
+
+        Run:
+            uvicorn main:app --reload
+    """
+    LIVE_RELOAD_HEADER = Script(f'{LIVE_RELOAD_SCRIPT}')
+    LIVE_RELOAD_ROUTE = WebSocketRoute("/live-reload", endpoint=live_reload_websocket)
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("hdrs", []).append(self.LIVE_RELOAD_HEADER)
```
comhar, line 57: Looks like `hdrs` and possibly `routes` can be tuples or lists. I'll add support for both types. 
Is there a preferred type? (e.g. if a list is received should we cast it to a tuple or vice-versa?)

jph00:
> I tested `FastHTMLWithLiveReload` with some local static files (js, css). To my pleasant surprise it handles this use-case as well 😄
> 
> Here's the setup I used for those tests:

You shouldn't need any of that, due to this:

https://github.com/AnswerDotAI/fasthtml/blob/main/fasthtml/core.py#L254

Here's an example of use:

https://github.com/AnswerDotAI/fasthtml/blob/main/examples/db_app.py#L16

> thoughts on what tests and docs should be included?

Might be hard to test automatically - but there's not much logic there, so maybe not needed?

I think just some basic docs in index.ipynb on how to use it would be sufficient. Sound OK?

fasthtml/live_reload.py
jph00: I guess lists are a bit easier to work with? -- but I don't have a strong preference.

Maybe for your code is this a little better?:

```python
kwargs["hdrs"] = [*kwargs.get("hdrs", ()), self.LIVE_RELOAD_HEADER]
```

comhar:
> I think just some basic docs in index.ipynb on how to use it would be sufficient. Sound OK?
Sounds good. Added a `Live Reloading` section.

> Might be hard to test automatically - but there's not much logic there, so maybe not needed?
True. As a sanity check I would like to add the basic test shown below. It doesn't cover everything but might catch some regressions in the future. I didn't include the test in the latest commit because it will fail (due to the routes issue). 

Let me know if you would prefer to commit the failing test now or include it in the PR for the routes fix. 

<details>

<summary>Test</summary>

```python
# create a regular app with a "/hi" route
app = FastHTML()
cli = TestClient(app)
rt = app.route

@rt("/hi")
def get(): return 'Hi there'

# re-create the same app using live reloading
lr_app = FastHTMLWithLiveReload()
lr_cli = TestClient(app)
lr_rt = lr_app.route

@lr_rt("/hi")
def get(): return 'Hi there'

# check that live reloading does NOT change the underlying behvaiour
test_eq(cli.get('/hi').text, "Hi there")
test_eq(cli.get('/hi').text, lr_cli.get('/hi').text)

# check that live reloading adds the expected route and header script
hdrs, routes = app.router.hdrs, app.routes
lr_hdrs, lr_routes = lr_app.router.hdrs, lr_app.routes

test_eq(len(lr_hdrs), len(hdrs)+1)
test_eq(lr_hdrs[0], lr_app.LIVE_RELOAD_HEADER)
test_eq(len(lr_routes), len(routes)+1)
test_eq(lr_routes[0], lr_app.LIVE_RELOAD_ROUTE)
```

</details>

jph00:
lgtm! thanks :D

jph00:
@comhar I've fixed the routes and added the test so please take a look and check it looks OK.

comhar:
> @comhar I've fixed the routes and added the test so please take a look and check it looks OK.

Sweet! I ran the latest commits locally and the live reloading is working as expected 👍 
The test looks good as well 👍

PR #30, opened 2024-07-03
Clarify railway docs

pydanny:
I know there's a command available, but I've done this enough now to want more clear instructions.


jph00:
Sorry @pydanny only just saw this (it's a good idea to at-mention me to ensure I get notified).

FYI we have `railway_deploy $proj_name --mount true|false` which makes this all a bit easier. You can see what it does here:

https://github.com/AnswerDotAI/fasthtml/blob/main/fasthtml/cli.py#L33

jph00:
Closing since IIUC this is now obsolete

PR #32, opened 2024-07-04
Fix ws example

Mihaiii:
Titled was not recognized

### DIFF
--- a/examples/basic_ws.py
+++ b/examples/basic_ws.py
@@ -1,5 +1,6 @@
 from asyncio import sleep
 from fasthtml.common import *
+from fasthtml.xtend import Titled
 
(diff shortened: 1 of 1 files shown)


johnowhitaker:
It works without this fix, but e.g. pylance doesn't pick it up in VSCode. @jph00  `xt_html` is not being found in xtend.py, and the three functions that use `@delegates(xt_html` in xtend.py aren't in xtend.pyi, so I think this is the root cause of this.

jph00:
The issue is that we needed to run `py2pyi` to update the .pyi file. I had assumed that pylance would be smart enough to pick up missing defs directly from the module, but it isn't -- so we'll need to re-run this whenever we update files with a .pyi version.

Fixed in main.

Issue #51, opened 2024-07-15
Better error messages when variables aren't passed into a view


pydanny:
Assuming this view:

```python
@rt('/')
def post(slug:str):
   return P(slug)
```

If a `slug` isn't passed in, and a default isn't specified, the 500 error returned isn't descriptive. Looks something like this:

```
... <snip a bunch of starlette stuff>
    response = await func(request)
  File "/Users/daniel.roy.greenfeld/projects/fasthtml/fasthtml/core.py", line 230, in _f
    resp = f(*wreq)
  File "/Users/daniel.roy.greenfeld/projects/fh-ui/fh_ui/main.py", line 26, in get
KeyError: <class 'inspect._empty'>
```

It would be nice if `KeyError: <class 'inspect._empty'>` was replace with something meaningful, both in terms of what is seen in the shell and what can be captured in the code.

PR #55, opened 2024-07-17
Raise error on missing params

pydanny:
For #51

TODOs:
- [x] Work with defaults set for params `def my_view(slug: str = None)`
- [ ] Use `test_fail` instead of a try/except block. 
- [ ] Fix or discuss failing test on `bodied` view, see image below in the "Failing Test item" block

## Tests 

Screenshot 2024-07-17 at 23 55 20

## Failing test item

Screenshot 2024-07-17 at 23 46 25

### When the test is uncommented here is the error it now generates

Screenshot 2024-07-17 at 23 57 02


jph00:
lgtm! i'll take a look at those other issues now.

PR #57, opened 2024-07-19
Updated html2xt to use unpacked dicts when attr keys are not valid python names

matdmiller:
Updated html2xt to use unpacked dicts when attr keys are not valid python names.

Also added example of using an unpacked dict to the explaining components nb.

Minimal Example:
```python
hl_md(html2xt_new('<div x-data="{open: false}" x-init="console.log(\'Initialized\')" @click="open = !open">Toggle</div>'))
```
to
```python
Div('Toggle', x_data='{open: false}', x_init="console.log('Initialized')", **{'@click':'open = !open'})
```

### DIFF
--- a/fasthtml/components.py
+++ b/fasthtml/components.py
@@ -120,6 +120,7 @@ def __getattr__(tag):
     return _f
 
 # %% ../nbs/01_components.ipynb 24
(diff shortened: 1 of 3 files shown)


matdmiller:
@jph00 just mentioning you to put this on your list in gh when you get a chance to look at it.

PR #58, opened 2024-07-19
Feature FastHTML by Example in the docs sidebar

audreyfeldroy:
This PR:

- Split FastHTML by Example into sections
- Reorder docs sidebar so it begins with FastHTML by Example
- Put rest of docs into Reference section a little lower in sidebar

The result:

- The docs lead with the part that feels friendliest to total beginners and most polished.
- It's hard to miss the detailed example walkthroughs. When I was starting out, I learned through the reference docs first, but I think I would have been more successful learning through the examples.
- This allows for expansion of the FastHTML by Example notebooks, which I'd like to explore.

Thoughts? Open to anything here. We could also launch with this and iterate later as other sections mature.


audreyfeldroy:
Hey @jph00 @ncoop57 review please 🙏

jph00:
Thank you for your 1st PR to FastHTML @audreyfeldroy ! 😄 

I realise in hindsight that I was answering your questions about how to split notebooks in Discord, without me actually asking *why* you were doing that -- which obviously if I was thinking clearly was the right time to have this discussion, rather than now after you've gone to all the trouble of creating this PR! I very much apologise for that oversight.

Could you say a bit more about your goals in splitting up the notebook into sections? AFAICT it comes with some significant downsides, which doesn't mean we can't do it -- just need to think through the pros and cons. Some downsides that come to mind:

- It makes it harder to just go thru the whole set of examples from beginning to end; we'd have to rely on users noticing a "go to next page" button or something. OTOH just continuing to scroll is a behaviour everyone is already used to
- Using ctrl-f to find things in a page is really convenient, to quickly jump through all the places something is mentioned. By splitting it up there's multiple places to look for examples
- Originally @johnowhitaker wrote this as much for an LLM to read as for a human to read; we'd need to add a manual step to create a combined version for an LLM, or for use with stuff like Claude Projects users would have to upload multiple files instead of just one
- The different sections aren't all standalone; i.e not every notebook is covering a well-defined topic which we can put a name to, to make it easy for users to know which link to click
- It makes the docs a bit harder to navigate -- once we've added the e2e tutorial from @ncoop57, and the introductions that I'm working on, there will be a lot of options for users to choose from; reducing the number of links might help make it less overwhelming
- There isn't a single table of contents to jump around as easily if it's split up; if it's in one file, Quarto can give us a nice ToC with scrollspy behaviour

As a personal preference (FWIW) I'm generally a fan of long-form pages, vs splitting things up, since I find I can get into the "reading zone" and zip thru the whole thing. E.g. each of our full book chapters are online as a single page: https://fastai.github.io/fastbook2e/intro.html .

jph00:
Here's a somewhat similar example from the nbdev docs in case it's helpful:

https://nbdev.fast.ai/tutorials/tutorial.html

We have quite a few separate sections in the docs, but the end-to-end walk-thru is just one page. There are then other pages for other specific deep dives.

audreyfeldroy:
Interesting. Whoa, point 3 changes my entire thinking around goals, writing style, ecosystem. Hey let's close this and I'll submit a new PR, something smaller.

jph00:
> Interesting. Whoa, point 3 changes my entire thinking around goals, writing style, ecosystem. Hey let's close this and I'll submit a new PR, something smaller.

Let me know if you'd like to talk about anything in person -- I'm always up for a chat!

PR #60, opened 2024-07-19
Clarify railway docs with fh_ commands

pydanny:
Combines #30 with instructions for the new script. Also more consistent with point-of-view.


jph00:
Thank you!

PR #73, opened 2024-07-22
Prefix ht2xt comment strings with '#' for easy identification

pydanny:
Note: This is an interim solution while we debug an issue with how bs4 detects comments. 

Converts:

```html
<div>
<!-- This is the first comment -->
<p>This is not a comment</p>
<!-- This is the second comment -->
</div>
```

To:

```python
Div(
    '# This is the first comment',
    P('This is not a comment'),
    '# This is the second comment'
)
```


jph00:
Can we make these actual comments, instead of strings?

pydanny:
> Can we make these actual comments, instead of strings?

BS4 isn't handling `isinstance(elm, Comment)` the way we expect. While we figure that one out this is an interim solution. Which I should probably mention in the main description.

jph00:
@pydanny I think you might have forgotten to export `_ht_com2py_com`.

jph00:
Closing since we're removing comments instead.

PR #95, opened 2024-07-30
Use standard indentation in README code examples

yuvadm:


jph00:
Thanks but we follow fast.ai style guide, not pep 8.

PR #108, opened 2024-07-30
Fix typos and simplify some phrasing on the ft component doc

willingc:


jph00:
Thank you for this excellent thorough editing Carol! :D

PR #117, opened 2024-07-31
Update ws example

mattlink:
Adds another endpoint to `examples/basic_ws.py` which illustrates how to read data from the websocket. 

It is not clear from the existing example (or docs, currently) that you can specify a `data` argument to the handler function to effectively get the contents of `ws.receive()`. There are other hidden arguments which seem to correspond to [these annotations](https://github.com/AnswerDotAI/fasthtml/blob/main/fasthtml/core.py#L199-L203)

### DIFF
--- a/examples/basic_ws.py
+++ b/examples/basic_ws.py
@@ -24,4 +24,8 @@ async def ws(msg:str, send):
     await sleep(2)
     return Div('Goodbye ' + msg, id=nid), mk_inp()
 
+@app.ws('/stream', conn=on_connect, disconn=on_disconnect)
+async def ws_stream(ws, data):
+    print(data)
+
 serve()


jph00:
Thanks - this is a good idea. I guess it also probably needs something on the page demonstrating its use?

jph00:
@mattlink just bumping this...

Kvit:
Another issue that I think should be covered in docs/examples is how to pass a variable from the web socket URL path. It is done via  `ws`  param.
Here is an example snippet 
```
from starlette.websockets import WebSocket

@rt.ws('/wscon/{session_id}', conn=on_connect, disconn=on_disconnect)
async def ws_func(ws:WebSocket, data, send):
    session_id = ws.path_params.get("session_id")
    #... rest of the code
```
NB: you can pass URL path param to `on_connect()`, `on_disconnect()` functions as well.

PR #130, opened 2024-07-31
use h2 for todo title in view

mathematicalmichael:
just a slightly more sensible default that helps differentiate the content of a todo from its title.


jph00:
Good idea!

Issue #132, opened 2024-07-31
Toasts don't work with bootstrap


vacmar01:
Apparently Toasts aren't visible when using Bootstrap. 

Bootstrap has its own `.toast` class and this might interfere with the `.toast` class provided by `setup_toasts()`. 

Or am I missing something here?

PR #133, opened 2024-07-31
changes toast names to avoid conflict with bootstrap

vacmar01:
See #132 

The `.toast` class of FastHTML (`setup_toasts()`) seems to be in conflict with the `.toast` class of bootstrap. 

Since Bootstrap is still very popular, it might be a good idea to namespace the class names to `.fh-toasts` etc. 

(I didn't run the nbdev commands, since the `toaster.py` doesn't seem to be created with nbdev. If that's wrong, I will change the pull request).


jph00:
Good idea! Fixes #132

PR #145, opened 2024-08-01
Add htmlkw to fast_app

ranzuh:
This PR adds the `htmlkw` parameter to the `fast_app` function, allowing it to pass this keyword argument to `FastHTML`.

After this fix, the following code will add `data-theme="light"` to the webpage's `<html>` tag:
```
app, rt = fast_app(htmlkw={"data-theme": "light"})
```


jph00:
Thank you for your first contribution to FastHTML! :D

PR #146, opened 2024-08-01
Fixed reloadAttempts variable of JS snippet in live_reload.py

pcky:
The counter for reload attempts in the JS snippet didn't work, throwing an error in Browser at live reload.


pydanny (approved):
Good catch! Was wondering why it was spotty sometimes!

PR #153, opened 2024-08-02
Use latest python packing standard pyproject.toml

gunungpw:
add `pyproject.toml`
add `build` (dev dependency) to make wheel and sdist

`pyproject.toml` follow the `setting.ini`

Next Action:
> Add docs to contribution: can be use uv, poetry, rye, pdm, also pip
- `git clone https://github.com/AnswerDotAI/fasthtml.git`
- `uv venv` # make virtual environment 
- `uv pip install -r pyproject.toml` # install project
- `source .venv/bin/active` # active the venv
- `python -m build` # build artifact to upload for pypi

> Remove? `setting.ini`, `setup.py`, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,3 +149,6 @@ checklink/cookies.txt
 .gitconfig
 
 _docs
+
+# UV lockfile
+uv.lock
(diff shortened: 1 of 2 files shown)


jph00:
Thank you for this contribution, but unfortunately this is not compatible with nbdev, which is what is used to write and build this library.

The nbdev project spent around a year trying to move to pyproject.toml but there was insufficient functionality in the toml-based approach to complete the transition.

PR #157, opened 2024-08-02
Fix typos in docstrings

### LINKED ISSUE #150: Typo in docstring in live_reload.py
https://github.com/AnswerDotAI/fasthtml/blob/3b992a001378149af6374e47e6f7caf24cd94e02/fasthtml/live_reload.py#L35

In the line above ^, there is a typo in the docstring. It should be "created" not "creaetd".

Nitpicky and minor issue but oh well 😅

dishb:
Fixes #150 as well as 1 other typo I found while looking through the codebase.

I changed the `oauth` and `xtend` notebooks. I also had to change `live_reload.py` directly, unsure if this is what I was supposed to do?

### DIFF
--- a/fasthtml/live_reload.py
+++ b/fasthtml/live_reload.py
@@ -32,10 +32,10 @@ class FastHTMLWithLiveReload(FastHTML):
     trigger a reload of both the server and browser window.
 
     How does it work?
-      - a websocket is creaetd at `/live-reload`
+      - a websocket is created at `/live-reload`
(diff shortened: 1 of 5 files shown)


johnowhitaker:
Thank you :) This looks good although there is some metadata changed in the extend notebook, maybe remove that file from this PR then I'll merge and fix that one myself?

dishb:
> Thank you :) This looks good although there is some metadata changed in the extend notebook, maybe remove that file from this PR then I'll merge and fix that one myself?

Should be good now, thanks!

PR #158, opened 2024-08-02
Add support for LaTeX formula rendering when rendering markdown

yym68686:
I added a few lines of code to support KaTeX formula rendering. Here is a test case:

```python
from fasthtml.common import *

hdrs = (
    MarkdownJS(), HighlightJS(langs=['python', 'javascript', 'html', 'css']),
    Link(rel='stylesheet', href='https://cdn.jsdelivr.net/npm/katex/dist/katex.min.css')
)

app, rt = fast_app(hdrs=hdrs)

content = """
Here are some _markdown_ elements.

- This is a list item
- This is another list item
- And this is a third list item

```python
print("Hello, World!")
```

$$
1 \\times 2 = 2
$$

$$1 \\times 2 = 2$$

$1 \\times 2 = 2$

**Fenced code blocks work here.**
"""

@rt('/')
def get(req):
    return Titled("Markdown rendering example", Div(content,cls="marked"))

if __name__ == "__main__":
    import uvicorn
    uvicorn.run(
        "__main__:app",
        host="0.0.0.0",
        port=5001,
        reload=True,
    )
```

### DIFF
--- a/fasthtml/js.py
+++ b/fasthtml/js.py
@@ -7,8 +7,14 @@ def  dark_media(css): return Style('@media (prefers-color-scheme:  dark) {%s}' %
 def MarkdownJS(sel='.marked'):
     src = """
 import { marked } from "https://cdn.jsdelivr.net/npm/marked/lib/marked.esm.js";
-import { proc_htmx} from "https://cdn.jsdelivr.net/gh/answerdotai/fasthtml-js/fasthtml.js";
-proc_htmx('%s', e => e.innerHTML = marked.parse(e.textContent));
+import { proc_htmx } from "https://cdn.jsdelivr.net/gh/answerdotai/fasthtml-js/fasthtml.js";
+import katex from "https://cdn.jsdelivr.net/npm/katex/dist/katex.mjs";
+
+const renderMath = tex => katex.renderToString(tex, {throwOnError: false, displayMode: false});
+
+proc_htmx('%s', e => {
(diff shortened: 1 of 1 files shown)


yym68686:
It seems GitHub automatically renders formulas. I'll paste a screenshot of the test code, which will make it clearer to see.

image

johnowhitaker:
Neat! @jph00 any objections to adding this to MarkdownJS?

jph00:
Let's add a bool param to enable/disable it, and default to false. Thanks!

jph00:
Oh also would be nice if start/end tags were configurable. E.g OpenAI models use different ones.

dzenny:
Could it be technically feasible and more useful to add direct mathjax support?
That way it would be possible to use tex-formulas not only in markdown, but in other contexts too, like buttons or other control elements.
All my attempts to add standard mathjax initiation script to hdrs, ftrs or Main have failed so far.

jph00:
> Could it be technically feasible and more useful to add direct mathjax support?

I think that would be an "addition to", not an "instead of"! :D Feel free to create a new issue showing what obstacles you're facing with mathjax, or drop by the discord (https://discord.gg/qcXvcxMhdP).

yym68686:
> Let's add a bool param to enable/disable it, and default to false. Thanks!

Hello, I have added a parameter to control whether to enable KaTeX rendering.

yym68686:
> Oh also would be nice if start/end tags were configurable. E.g OpenAI models use different ones.

I have already added custom start/end tags to the MarkdownJS function.

yym68686:
This is the new test script: https://pb.yym68686.top/ywjD?lang=python

jph00:
Thank you!

PR #159, opened 2024-08-02
Update Documentation: Replace default_hdrs=False with pico=False in 'CSS Files and Inline Styles' section

Finnegan1:
The documentation in the "Web Devs Quickstart" section "CSS Files and Inline Styles" advises to use `default_hdrs=False` to exclude pico.min.css. This does the job, but in my opinion is misleading with the explanation below. `default_hdrs=False` removes all headers, including those for HTMX. In my opinion, `pico=False` would fit much better into the example and the explanation below.

^At least according to my understanding, but happy to be wrong :) 

screenshot


jph00:
Well spotted! :D

PR #169, opened 2024-08-03
Changes to run prep command on commit automatically and updated contribution.md to help beginners to setup project.

Rahulbeniwal26119:
- This PR Introduces a pre-commit hook to run prep command automatically on commit. So there is no need to run it manually and no need to set it as an alias.
- This PR also adds ipdb as a dev dependency with pre-commit.
- This PR also adds more hooks to strip white spaces, check for debug statements, and others.
- This PR also adds some more instructions in the get started section to help new contributors set up project easily in case they never worked with setup files before.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+repos:
+  - repo: local
+    hooks:
+        - id: prep
+          name: clean Jupyter notebooks before committing
+          entry: bash -c "nbdev_export && nbdev_clean && nbdev_trust"
+          language: system
+          pass_filenames: false
+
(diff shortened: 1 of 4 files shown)


jph00:
Thanks for this! FYI `python3 setup.py install` shouldn't be used directly; instead use `pip install -e .`.

For the commit stuff, please take a look at the nbdev hooks in the nbdev project. I think the best place to add most of your PR would be over there, if you find things the existing hooks don't do to your satisfaction.

Rahulbeniwal26119:
Thanks, @jph00 be I will raise a PR there. :+1:

PR #170, opened 2024-08-03
Explanation around string handling and escaping added to the document…

MartinEls:
Addition to the web dev documentation to address some confusion around string handling and the escaping of HTML tags. https://x.com/jeremyphoward/status/1819443235775041657


jph00:
Thank you :D

PR #185, opened 2024-08-04
Make fill_form set selected option

ostwilkens:
Make fill_form set "selected" attribute on Options within Selects, so this is enough:
```python
Select(
    *[Option(x[0], value=x[1]) for x in RARITIES],
    id="rarity",
),
```
instead of
```python
Select(
    *[Option(x[0], value=x[1], selected=item.slot is x[1]) for x in RARITIES],
    id="rarity",
),
```

### DIFF
--- a/fasthtml/components.py
+++ b/fasthtml/components.py
@@ -82,6 +82,10 @@ def _fill_item(item, obj):
                 else: attr.pop('checked', '')
             else: attr['value'] = val
         if tag=='textarea': cs=(val,)
+        if tag=='select':
+            for o in cs:
+                if o[0]=='option':
+                    if o[2].get('value', '')==val: o[2]['selected'] = '1'
     return FT(tag,cs,attr)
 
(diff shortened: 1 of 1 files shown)


fasthtml/components.py
```diff
@@ -82,6 +82,10 @@ def _fill_item(item, obj):
                 else: attr.pop('checked', '')
             else: attr['value'] = val
         if tag=='textarea': cs=(val,)
+        if tag=='select':
+            for o in cs:
+                if o[0]=='option':
+                    if o[2].get('value', '')==val: o[2]['selected'] = '1'
```
jph00, line 88: You could combine this with the previous line using `and` IIUC.

jph00 (changes requested):
Many thanks -- good idea! This is an nbdev project, so please edit the source notebook then export it to the .py file.

jph00:
I'll pop this into a notebook.

Issue #131, opened 2024-07-31
CRUD Example


GirishKumarSharma:
Is it possible to please give me a complete CRUD example of using database i.e. Add/Edit/Delete/Next/Previous options in table rows by having Checkbox, Combo Box, Textbox, Option Button, Toggle Button, Command Button etc. and almost all most used controls with Comments and clarifications, so that it may increase its usages.

Kishlay-notabot:
I've been reading [this](https://github.com/AnswerDotAI/fasthtml/blob/main/examples/adv_app.py) example provided and it has nicely documented code. I'd get started with a crud app if I get time, till then I recommend you guys to read the code on the link above. 
Thanks

rajbhoyar729:
hi  i want to work on this issue

jph00:
Actually we've got one from @rohitmidha23 now! I'm just trying to convince him to pop it in a repo and give it a readme :D

PR #188, opened 2024-08-04
Example: Simple CRUD App - #131

RohitMidha23:
Example App to demonstrate CRUD functionality - for issue #131

### DIFF
--- a/examples/crud_app.py
+++ b/examples/crud_app.py
@@ -1,0 +1,161 @@
+###
+# Walkthrough of a FastHTML CRUD application for managing books
+###
+
+
+# Importing from `fasthtml.common` brings together key parts of fastcore, starlette, fastlite, and fasthtml.
+from fasthtml.common import *
+from dataclasses import dataclass
+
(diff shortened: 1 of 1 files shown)


jph00:
Good job @RohitMidha23! I think rather than putting this in the main repo, it would be more effective to showcase it as a community contribution -- that way you can create a readme or blog post showing what the finished app looks like, how it works, etc. Would that be OK? If you did that we'd be delighted to highlight it!

RohitMidha23:
@jph00 this particular example feels too simple to have as a separate blog post imo. Can work on a more complex one maybe but can’t commit as yet. If you have any ideas for that one, let me know!

jph00:
I don't agree! I think it's great :D But up to you of course...

GirishKumarSharma:
I think @RohitMidha23 is right. A more extended example will include Date, Short Notes, Boolean and Image data types by including some more common controls like Combo Box, Check box, Option button, Toggle button etc. So that it may increase its usability as well.

jph00:
Sure -- feel free to have a go at that and if you share it here I'll take a
look.

PR #201, opened 2024-08-05
Fix bug: fix the bug where multi-line formulas were incorrectly rende…

yym68686:
I fixed the bug where inline equations were incorrectly rendered as display equations. Here is the test code: https://pb.yym68686.top/NW6T?lang=python


jph00:
Great!

PR #202, opened 2024-08-05
Handle mismatches between existing db and schema

ncoop57:
This PR attempts to detect and give a user option to how they want to handle a schema mismatch in the database and fast_app function. Right now a user can just specify to overwrite the current table. However, it doesn't currently handle types.


ncoop57:
@jph00 do you know of any fancy way to compare dataclass type hints? Could try looking at manual validation like this SO answer: https://stackoverflow.com/questions/50563546/validating-detailed-types-in-python-dataclasses, but curious if you know of an easier way that doesn't require having pydantic as a dep

jph00:
fastlite has a `transform` method (actually from sqlite-utils) -- I'd suggest looking into whether that has a solution (i.e just update the DB so it matches the class).

ncoop57:
omg that was so much easier 😅 Thanks @jph00! I think it is ready to go now!

PR #206, opened 2024-08-06
Add section on FastHTML hx_ attributes to README

ak4zh:
I noticed that there wasn't a single comprehensive resource listing all the hx_ attributes used in FastHTML. To improve accessibility and provide quick reference for developers, I have added a section to the README.


jph00:
I'd rather just link folks to the HTMX docs, since there's a lot more than what's in that list. Thanks for the PR though!

PR #213, opened 2024-08-06
Wrap Starlette's session in an AttrDict

jbellis:
Testing status:

fasthtml: 
- nbdev_test passes except oauth, which also fails without this patch (`No such file or directory: '../examples/gh_secret.env'`)

manually testing fasthtml-examples using sessions:
- todos2: works, including when modified to use attribute syntax
- image_app_session_credits: skipped (requires giving Replicate my billing info)
- data_spot_check: works
- 03_pictionary: `No such file or directory: 'multiplayer.css'` (same error without this patch)


jbellis:
Also confirmed that home and about work, although neither appears to use sessions.

jph00:
Thank you! PS: tag me for review, or at-mention me, so I don't miss PRs in the future.

jph00:
Oh and also - thank you for the careful testing :D

jbellis:
Thanks, and will do!

PR #221, opened 2024-08-08
Use re.fullmatch for Beforeware

justbur:
The docs, specifically [here](https://docs.fastht.ml/tutorials/quickstart_for_web_devs.html#authentication-and-authorization), suggest using

```python
beforeware = Beforeware(
    user_auth_before,
    skip=[r'/favicon\.ico', r'/static/.*', r'.*\.css', r'.*\.js', '/login', '/']
)

app, rt = fast_app(before=beforeware)
```

to set up user auth, but the '/' route in skip currently matches all routes. It seems the intention was to match the entire route, hence the change to re.fullmatch in this commit.


jph00:
Many thanks!

Issue #229, opened 2024-08-09
`fh_railway_deploy` doesn't work for new app / doesn't check for project name


gautam-e:
## Issue
When I run `fh_railway_deploy NEW_APP` to deploy a new app to a new project, it says that the project already exists (even though it doesn't!).

## What I checked
When I run `railway_status` it shows an existing project that I have on railway:
```
Project: OLD_APP
Environment: production
Service: OLD_APP
```
But I want to deploy my new app to a new project!
After looking at the `railway_deploy` function, I think I found the problem. It doesn't check for the Project name.

## Proposal
I think I have figured out the solution and have modified `railway_deploy` function in the `09-cli.ipynb` notebook as well. 
I can make a PR pretty quickly if someone confirms that this is indeed a problem that needs such a solution (just to be sure I'm not doing something silly here!).

PR #230, opened 2024-08-09
railway deploy doesn't check for project name

gautam-e:
This address #229


jph00:
Good job - thank you!

PR #232, opened 2024-08-09
add basic typing (and small fix for _find_wsp)

mjbommar:
I understand that you might not be a huge fan of typing in Python, @jph00, but as someone trying to learn how fasthml is designed and works, I found myself frequently looking up the return types anyway (especially in the absence of docstrings).

Since it was useful for my own learning as I went through the source, I added type annotations where I was confident enough.  No hurt feelings if you don't want to merge (although there were some scattered type annotations already, so it seems like not such a dogmatic position 😀).

There was also a small error in `_find_wsp` which I corrected while annotating and manually testing all examples (https://github.com/AnswerDotAI/fasthtml/commit/551745ccfae5fb47c8a407de8b16b89170bc7344).  You might cherrypick on your own commit since it is such a small fix to PR.


jph00:
Thanks @mjbommar ! The bad news, I'm afraid, is that this is an nbdev project, so the source code is the notebooks, not the .py files. The .py files are auto-generated. So I can't merge this since it just be overwritten by the notebooks anyway.

I've grabbed the ws header fix and will push that -- well spotted!

In order to explore the lib, I recommend using a notebook, such as running the source notebooks themselves. That way you can directly inspect the actual types created. Most functions already have in-notebook examples allowing you to do this already.

mjbommar:
@jph00 oops!  I see nbdev in the contributing guide now :smile:

PR #233, opened 2024-08-09
Fast html mention

Isaac-Flath:
Added FastHTML.Gallery in third-party section.  Should it be here or in the "official" section in the future?

@jph00


jph00:
Thanks!

PR #256, opened 2024-08-11
Update instructions for tags with invalid symbols

Isaac-Flath:
@jph00  Update instructions for tags with invalid symbols to use a dictionary.


jph00:
Much better!

Issue #226, opened 2024-08-08
Session can't encode objects it can decode


jbellis:
I noticed it with UUIDs, other types may also be affected.
```
from uuid import UUID
from fasthtml.common import *

app = FastHTML()

@app.get("/")
def index(session, item_id: UUID):
    session['item_id'] = item_id

serve()
```
```
curl -X GET "localhost:5001/?item_id=36621c53-55c3-11ef-b14b-c45ab1ddc9ad"
```
Result:
```
TypeError: Object of type UUID is not JSON serializable
```

palomino79:
This is less an issue with FastHTML and more an issue with Starlette, which FastHTML is built on top of. Specifically, FastHTML overrides Starlette's SessionMiddleware in order to wrap the initial session data in an AttrDict. On sending session data, Starlette serializes the session object via a call to `json.dumps`, which examines the type of object it's serializing and raises a TypeError if it finds something in it that is not one of the object types it knows how to work with. As such, I see two practical solutions to this problem. 

1. Require developers to cast whatever object they want to persist in session data to something they know can be handled by `json.dumps`. In this case, that would be simply casting the UUID to a string. Or...
2. Overriding the rest of the SessionMiddleware's  `__call__` method to integrate a more flexible serialization function inside of `send_wrapper` that would attempt to convert the objects it was passed into known serializable data types, before ultimately being passed to `json.dumps` and proceeding with the rest of the expected logic of that function.

If a project maintainer could let me know whether or not the second option is in line with the general design philosophy of this project (or, more generally, if such a change would be desirable), I'd be willing to attempt the necessary changes to accommodate the second option and then open a PR for it.

jph00:
> If a project maintainer could let me know whether or not the second option is in line with the general design philosophy of this project (or, more generally, if such a change would be desirable), I'd be willing to attempt the necessary changes to accommodate the second option and then open a PR for it.

We do have a (slightly) custom `_SessionMiddleware` now, so there's certainly room to experiment here. Maybe give it a go and see if you're able to come up with something that feels simple, consistent, and intuitive? (If we can't, I'm fine telling folks to just serialize things manually.)

PR #262, opened 2024-08-11
Custom serialization for session data

palomino79:
Associated issue number: #226. In that issue, I list a couple of options for dealing with certain objects added to session data not serializing.

The reason this is happening:
FastHTML overrides Starlette's SessionMiddleware in order to wrap the initial session data in an AttrDict. On sending session data, Starlette serializes the session object via a call to json.dumps, which examines the type of object it's serializing and raises a TypeError if it finds something in it that is not one of the object types it knows how to work with. As such, I see two practical solutions to this problem.

The solution:
Overriding the rest of the SessionMiddleware's `__call__` method to integrate a more flexible serialization function inside of send_wrapper that would attempt to convert the objects it was passed into known serializable data types, before ultimately being passed to json.dumps and proceeding with the rest of the expected logic of that function.

Additional notes:
I ran the specified combination of nbdev_export && nbdev_clean && nbdev_trust prior to committing, but it seems like there's some traceback data that is a part of the commit towards the bottom. It's possible I did something wrong there. If so, if someone can direct me towards what I should've done right, then I can refactor as necessary. Also, I have tried to follow along with the expected contributing guidelines. I'm hoping I wrote the unit test per the maintainers' expectation, but if not, I'm happy to revise it.


palomino79:
As an additional point of consideration, I could see it being worthwhile to iterate through the keys and values of a dictionary stored in the session, as well as any lists, and attempting to convert those values to serializable formats. This would help to account for users doing things like, for example, storing a list of UUIDs in the session object, or storing a dictionary where the keys and/or values were also UUIDs. If that seems like a valid use case the application should account for, I could do a quick refactor of the serializer function to include that.

jph00:
This is too heavyweight for my liking -- even although it's cool! How about if we make the session backend pluggable, then this more fancy one could be an extension people could choose?

palomino79:
@jph00 I definitely agree that just ripping out the full Starlette method implementation and forcibly injecting additional parsing logic into it is a bit invasive and could lead to unexpected behavior under certain circumstances. That said, could you maybe provide a bit more detail in what you think would be a good approach here? Are you thinking maybe preserving Starlette's implementation by default and then allowing a user to just pass a flag into the FastHTML initializer indicating that they'd like fault-tolerant session serialization (a simple if/else statement, really, inside the `__call__` method controlling its behavior), or perhaps actually allowing the user access to a completely distinct, fault-tolerant SessionMiddleware class that they could import and provide as an argument to the FastHTML object? Lots of potential approaches here, really. Just trying to gauge what you think would be a good balance between user control, simplicity, and explicitness.

palomino79:
I've updated this branch to add a separate session middleware file, with an example test for how it can be used. It's a bit awkward to have to directly append the middleware to FastHTML's user_middleware list, but the way that Starlette handles user middleware is to aggressively prepend the middleware list with them. That means if you want to add something that runs prior to SessionMiddleware on the way back up the middleware stack, you have to force it to the end of said list. I considered creating an additional method in the FastHTML class to wrap the append to the middleware list, but that seemed a bit invasive for accommodating the requirements of a single type of middleware.

jph00:
i added a `sess_cls` param to fasthtml yesterday -- how about making your serialized session an external project, which people got opt in to using `sess_cls`?

palomino79:
@jph00 Alright, so, when you say an external project, you're saying that this session middleware class just shouldn't be a part of FastHTML all at all? In that case, should I go ahead and just close this PR, then?

jph00:
Yes, if you think it will work OK to make it an external project, we should do that -- and if you do, please let us know so we can link to it from our docs.

palomino79:
@jph00 Alright, if it's not too much trouble for the maintainers of FastHTML, I'll start a project for fasthtml-middlewares (I'd imagine something akin to how Flask handles additive tooling with things like flask-smorest, flask-wtf etc.). I think something that purely exists to handle session middlewares might be a bit limited, as there's only so much you can do with the session object, but a larger project to aggregate useful middlewares for FastHTML might be pretty helpful for folks going forward.

jph00:
Sounds great! Just let us know how we can best support you as you get that
going. :D

PR #264, opened 2024-08-12
added functionlity to load style from file with chache on modified date

Rahulbeniwal26119:
- This change will allow a developer to keep the CSS files inside the proper CSS file instead of defining them inside the .py file.
- I have also added caching to the CSS file. The cache will reset if the file is modified.

If we already have this functional, please let me know and I would be okay to decline this PR.
Thanks


jph00:
We recommend using `File` or `Link` for putting CSS in external files -- let us know if that doesn't meet your needs. (PS: you can use fastcore's `flexicache` for mtime-based caching).

PR #279, opened 2024-08-13
fix: fix a bunch of typos using codespell (and manual review)

corneliusroemer:
Fixing a bunch of typos in a systematic way by configuring and running codespell

This PR contains just the typo fixes, the .codespellrc is in a separate PR in case you don't want the extra file in the repo.

PR #280, opened 2024-08-13
chore(ci): add .codespellrc to simplify checking for typos

corneliusroemer:
Running `codespell` with this config results in no false positives and fixes the typos as in #279


jph00:
Thanks!

PR #294, opened 2024-08-14
Document the MiniDataAPI Specification

pydanny:
- [x] Add `contains` explanation
- [x] Improve `xtra` docs
- [x] `transform` is a kw argument in `db.create` that performs necessary data structure migrations, but not data migrations. It isn't working. See https://github.com/AnswerDotAI/fastlite/issues/21
- [x] `__call__` takes a where clause, so that is dependant on the DB implementation
- [x] `__call__` takes a limit
- [x] `__call__` takes an offset


nbs/explains/minidataapi.ipynb
```diff
@@ -0,0 +1,516 @@
+{
```
jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 5 Line 1_
```python
db = database('data/minidataapi.db')
```
maybe use in-memory for this nb?

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 6 Line 1_
```markdown
## Create database tables
```
This section should be removed - it's not part of the spec

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 9 Line 3_
```markdown
## .insert()

Add a new record to the database. Must accept Python dictionaries, dataclasses, and standard classes. Returns an instance of the new record.
```
Have you checked whether fastsql supports dicts? I can't recall if it does. If there's no tests of that in fastsql, but it does work, please add tests/examples.

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 10 Line 1_
```python
# Add records to the users table
```
If we support both dicts and objects, we should show examples of both. Most examples should use objects, since that's what I think people should be using most of the time.

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 10 Line 4_
```python
# Add records to the users table
users.insert({'name': 'Braden', 'email': 'b@example.com', 'year_started': 2018})
users.insert({'name': 'Alma', 'email': 'a@example.com', 'year_started': 2019})
users.insert({'name': 'Charlie', 'email': 'c@example.com', 'year_started': 2018})
```
It's unusual to just show the last inserted object -- or at least without mentioning that. Maybe instead for the last row say `user = ...` and then add a new cell underneath with just `user` in it, and a explanation saying what this now contains.

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 11 Line 4_
```python
# Now the todos table
todos.insert(dict(title='Write MiniDataAPI spec', status='open', name='Braden'))
todos.insert(dict(title='Implement SSE in FastHTML', status='open', name='Alma'))
todos.insert(dict(title='Launch FastHTML', status='closed', name='Charlie'))
```
This is the main interesting one to point out -- that the returned todo has the updated id in it. It's not just the object we passed in. Also the dict has been converted to the class. These are the important details to not only show, but also explain why this is important and how it's used in fasthtml (for example).

i.e we need to explain *why* the spec has these particular details.

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 15 Line 1_
```python
try:
```
I don't normally mind about formatting for code much, but for sample docs I care more -- so let's follow fast.ai style here and put the `try` body on the same line.

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 12 Line 3_
```markdown
## Square bracket search []

Get a single record by entering a primary key into a table object within square brackets. 
```
Can also be multiple values for composite keys.

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 18 Line 1_
```markdown
## Circle bracket search ()
```
I've not heard them called circle brackets before - is that common? "Parentheses" is the term I'm more familiar with

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 19 Line 1_
```python
for user in users():
```
Same line here too -- or better still just say `users()` and that's it. No need for `print` or a loop afaict

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 23 Line 1_
```python
for user in users(where="year_started=2019"):
```
Let's use a named parameter here to demonstrate that. Also the spec includes a number of other params, so show those too. (see fastsql or fastlite for details)

jph00, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 32 Line 3_
```markdown
## .xtra()

The xtra action adds a filter to queries and DDL statements. This makes it easier to limit users (or other objects) access to only things for which they have permission. 
```
Show ddl examples too and explain the details of how they work.

jph00 (changes requested):

nbs/explains/minidataapi.ipynb
pydanny: fastsql doesn't support the use of adding records via dicts. I'll add that capability in that library.

pydanny: Fastlite supports Dataclasses and dicts, but not regular classes.

pydanny: I went with what you said, I thought it was a nifty phrase that no one else used. 🤣

nbs/explains/minidataapi.ipynb
```diff
@@ -0,0 +1,889 @@
...
```
nwpulverizer, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 41 Line 1_
```python
users()
```
Is this supposed to have some parameter like ```users(sort="name")```?

nwpulverizer, line 1: _Commented on notebook `nbs/explains/minidataapi.ipynb` Cell 66 Line 1_
```python
todos.xtra(name='Alma')
```
```todos.xtra(name='Charlie')``` I think is what you mean to have here

pydanny:
This error happens if `__contains__ is done with a tuple:

Screenshot 2024-08-21 at 16 36 13

nbs/explains/minidataapi.ipynb
pydanny: I just spotted this comment. 🤦🏻 

For DDL examples, can you explain or provide an example of what you are looking for?

jph00: Specially, show how `.xtra` changes `insert`, `update`, and `delete`. (Check the fastlite docs/examples and source for details.)

pydanny: Thanks for the coaching! I'm delighted to finally give it a proper go. 😄

jph00:
@pydanny also note that you would not be able to change the username on a to-do when xtra is in effect. Also, it's a little confusing IMO that the doc refers to `.contains`, but that's not a method we define. Maybe say `in`? (and maybe also note that this behavior is automatic in Python as long as you have defined `__getitem__`?)

jph00:
@pydanny do you think this is ready to merge now?

pydanny:
@jph00 Barring your suggested change to `xtra| i think it is ready for merge. 🙂

pydanny:
@jph00 I think this is ready to merge. Any other changes you think should go in?

jph00:
Excellent! Many thanks! :D

Issue #231, opened 2024-08-09
Beforeware cannot be easily associated with a route.


sergue1:
As of now, Beforeware is used in application scope, e.g.: 
```python
    app, _ = fast_app(before=Beforeware(authentication))
```

It can be useful to associate it with a route, e.g. to check authorization. Currently, this workaround works:
```python
    app.router.add_route("/", homepage)
    app.router.add_route("/users", users)
    app.router.routes[-1].before = [Beforeware(authentication), Beforeware(require_role("admin"))]
```

Note, we cannot just `app.router.routes[-1].before.append(Beforeware(require_role("admin")))`, because it gets applied globally then.

Besides, it would be helpful for large applications to organize routes in groups with a given prefix and Beforeware to avoid duplications.

PR #295, opened 2024-08-14
feat: (#231) Add route-specific Beforeware support

khoaHyh:
Addresses: https://github.com/AnswerDotAI/fasthtml/issues/231

## Summary

- Modified `RouteX` and `RouterX` to handle route-level Beforeware
- Maintained existing Beforeware functionality
- Added tests for various Beforeware scenarios
  - ensure app-level beforeware still runs
  - ensure route-specific beforeware is ran for the endpoint it is associated to
  - ensure Beforeware association is supported through different ways of defining routes


jph00:
Thanks for this PR! I'm open to merging this, but first I want to see whether there's other approaches that might work. Have you looked into whether using a simple decorator would provide the sample thing that you're wanting to use route-specific Before for?

If you have some examples of where that isn't a good solution, could you share a couple here please?

khoaHyh:
@jph00 Thanks for the feedback, I did consider a decorator approach initially but chose the parameter method for a straight forward solution with minimal changes. I don't have any specific examples but I'm curious to learn about the preference for decorators.

I'm open to exploring a decorator-based approach if you think it's more appropriate. Here's my try at using a simple decorator:

```python
@patch
def route(self:FastHTML, path:str=None, methods=None, name=None, include_in_schema=True):
    "Add a route at `path`; the function name is the default method"
    pathstr = None if callable(path) else path
    def f(func):
        n,fn,p = name,func.__name__,pathstr
        assert path or (fn not in _verbs), "Must provide a path when using http verb-based function name"
        if methods: m = [methods] if isinstance(methods,str) else methods
        else: m = [fn] if fn in _verbs else ['get'] if fn=='index' else ['post']
        if not n: n = fn
        if not p: p = '/'+('' if fn=='index' else fn)
        ## Changes below
        beforeware = getattr(func, '__beforeware__', [])
        self.router.add_route(p, func, methods=m, name=n, include_in_schema=include_in_schema, before=beforeware)
        ## Change above
        lf = _mk_locfunc(func, p)
        lf.__routename__ = n
        return lf
    return f(path) if callable(path) else f

all_meths = 'get post put delete patch head trace options'.split()
for o in all_meths: setattr(FastHTML, o, partialmethod(FastHTML.route, methods=o))

# new decorator
def before(*funcs):
    def decorator(route_func):
        if not hasattr(route_func, '__beforeware__'):
            route_func.__beforeware__ = []
        route_func.__beforeware__.extend(funcs)
        return route_func
    return decorator


# Testing Beforeware association to routes
counter = { "app": 0, "route": 0, "auth": 0 }

def app_beforeware():
    counter["app"] += 1

def route_beforeware():
    counter["route"] += 1

def authentication():
    counter["auth"] += 1

def users():
    return "Users Page"

app,cli,rt = get_cli(FastHTML(before=Beforeware(app_beforeware)))

@app.route("/")
def get():
    return "Hello app-level beforeware"

@app.route("/route-level")
@before(route_beforeware) # <-- associating beforeware specifically to a route via decorator
def get():
    return "Hello route-level beforeware"

@rt("/admin") 
@before(Beforeware(route_beforeware), Beforeware(authentication))
def get():
    return "Admin page"

app.router.add_route("/users", users, before=[Beforeware(route_beforeware), Beforeware(authentication)])

app_level_response = cli.get("/")
assert app_level_response.status_code == 200 and app_level_response.text == "Hello app-level beforeware" and counter == { "app": 1, "route": 0, "auth": 0 }

route_level_response= cli.get("/route-level")
assert route_level_response.status_code == 200 and route_level_response.text == "Hello route-level beforeware" and counter == { "app": 2, "route": 1, "auth": 0 }

admin_response= cli.get("/admin")
assert admin_response.status_code == 200 and admin_response.text == "Admin page" and counter == { "app": 3, "route": 2, "auth": 1 }

users_response = cli.get("/users")
assert users_response.status_code == 200 and users_response.text == "Users Page" and counter == { "app": 4, "route": 3, "auth": 2 }
```

I've kept the `before` parameter in `add_route` but removed it from `route` so if we want to assign Beforeware to a route you'll have to use `@before` instead of passing it in as a param. You are still able to pass in Beforeware via `app.router.add_route('/users', usersHandler, before=[Beforeware(auth)])`

jph00:
Apologies, I didn't clearly communicate what I meant. My question wasn't
about your implementation, but about whether we want route-specific
beforeware at all. Instead, could users just decorate their functions to
achieve the same purpose? This is how the Flask community tends to do
things, for instance. Does that make sense?

khoaHyh:
Thanks for clarifying. I see your point now and agree that using Python decorators is a simpler and more flexible approach. I think we should close this PR and instead document the decorator-based method for users. Perhaps we could include a test as documentation, what do you think?

jph00:
I love that idea, thank you! Are you open to having a try at a doc PR?

khoaHyh:
@jph00 Yep! PR is here 👉🏻 https://github.com/AnswerDotAI/fasthtml/pull/316

Issue #286, opened 2024-08-13
Windows - Uvicorn fails to properly shutdown process when using ctrl c from terminal


brendanboyle87:
Hi all,

Thanks for putting together such a great framework! I just started learning and encountered an issue in `nbs/tutorials/by_example.ipynb` while using `uvicorn app:app --reload` to run my hello world app on Windows. As I went through the examples, I noticed that sometimes my webpage was still being served even when the server was shut down in my terminal (and server restarts would not pick up my changes). It looks like this issue is related to [this GitHub issue](https://github.com/encode/uvicorn/issues/1972).

I switched to using WSL, and everything worked fine. Perhaps the documentation could be updated to suggest this workaround for Windows users? I can open a PR for the update but wanted to check if this makes sense to the maintainers first.

jph00:
Sure, we'd be very happy to have such a doc update to suggest WSL. Note that docs live inside the notebooks.

PR #299, opened 2024-08-14
update server run documentation to recommend WSL for Windows users

brendanboyle87:
#286  Updated docs to include recommendation to use WSL on Windows

WSL update pic


jph00:
Thanks!

Issue #247, opened 2024-08-10
Using `sess` as the first argument is not setting the values in the session


ramonvg:
First of all, congratulations! I'm very hyped about this project.

I've been playing around for a short time, but I couldn't figure out what I was doing wrong. After debugging a bit, I came up with this small working example. You basically get a login page where, if you type `admin`, you are logged in.

```python
from fasthtml.common import *

login_redir = RedirectResponse("/login", status_code=303)


app = FastHTML()
rt = app.route


@rt("/login")
def get():
    frm = Form(
        Input(id="name", placeholder="Name", autofocus=True),
        Button("login"),
        action="/login",
        method="post",
    )
    return Titled("Login", frm)


@rt("/login")
def post(name: str, sess):
    if name != "admin":
        return login_redir

    sess["auth"] = name
    return RedirectResponse("/", status_code=303)


@rt("/logout")
def get(sess):
    del sess["auth"]
    return login_redir


@rt("/")
def get(sess):
    if "auth" in sess:
        content = Div(
            P(f"Hello, {sess['auth']}!"),
            A("Logout", href="/logout"),
        )
    else:
        return login_redir
    return Titled("Home", content)


serve()
```

The bug occurs when you flip the positions of `name` and `sess` in the `post` for `/login`.

```python
@rt("/login")
def post(sess, name: str):
    ...
```

It seems `sess` doesn't like to be in the first position, since when you add another parameter and place it in the middle, it works as expected. It also fails if you move it to the first position. This example works.

```python
@rt("/login")
def post(name: str, sess, password: str):
    ...
```

This one fails:

```python
@rt("/login")
def post(sess, name: str, password: str):
    ...
```

Tested on `python-fasthtml==0.3.0`

Rahulbeniwal26119:
@ramonvg will look into it.

velaia:
I've tried to reproduce the issue with `sess` (or `session`) in the first position but couldn't: the sample entry I put into the session stays.

My issue is slightly different in that the toasts I add somehow don't get rendered when using a RedirectResponse despite being in the session. I'm investigating and might create a new issue.

vacmar01:
@velaia, I have exactly the same problem with toasts not rendering after redirecting.

pydanny:
This is a test case I wrote in core that demonstrates the failure:

```python
# %% ../nbs/api/00_core.ipynb
@rt("/sess-delete")
def delete(sess):
    sess.pop('authx', None)
[...]

PR #308, opened 2024-08-15
Test for session argument order

pydanny:
This is tests for replicating #247, which fails if tested locally on my laptop. Unfortunately, these don't replicate the error reported or proven.


bigcharless:
> This is tests for replicating #247, which fails if tested locally on my laptop. Unfortunately, these don't replicate the error reported or proven.

Hello, this reproduces the issue https://replit.com/@charlesrongione/Redirect-clears-session?v=1

PR #316, opened 2024-08-16
docs(test): #231 Add example for decorator-based route-specific functionality

khoaHyh:
Addresses https://github.com/AnswerDotAI/fasthtml/issues/231
Closes https://github.com/AnswerDotAI/fasthtml/pull/295

### Changes

Added a test as documentation for decorator-based method to add route-specific functionality


jph00:
Thanks this is great! I actually think this would be more useful to folks if you added it to this page (or whatever page you think it best fits): https://docs.fastht.ml/ref/defining_xt_component.html. Instead of just a list of tests, how about adding some browse about what you're showing here, how/why it works, and what it's useful for?

khoaHyh:
@jph00 I opted to add documentation to the [Routes page](https://docs.fastht.ml/explains/routes.html) instead because I think it might interest developers to customize specific behavior pertaining to routes. The API doesn't enable passing things like `Beforeware` to the route function but it might be helpful to know there's some flexibility with routes.

jph00:
That's a great approach - thank you for your patience and tenacity on this!

PR #378, opened 2024-08-29
Make improvements to `FastHTML By Example`

hamelsmu:
Small PR as I'm working my way carefully through this document.  I suspect I will make multiple PRs onto this document. 

Changes made:

1. Use `serve()` instead of Uvicorn
2. Include links in places where there should be links
2. Make some of the language more succint (less is more)
3. Remove the bit about this doc serving as context because there is now `llm-ctx.txt` and friends
4. Update for new `@rt` decorator, and add a tabset panel so people can see the difference
5. Clarify some of the language assuming people are new to web development.
6. Move all `Asides` into Quarto callouts

Im only 25% of the way through just though I would open an early PR.  Don't want to make a big PR if these changes are not aligned with what you want to do!

>[!Important]
> I have generated [a preview link for the page I changed](https://66d0c49fad8823743fde7a8e--fasthtml-preview.netlify.app/tutorials/by_example) to make reviewing easier on Netlify

cc: @johnowhitaker @jph00

### DIFF
--- a/nbs/tutorials/by_example.ipynb
+++ b/nbs/tutorials/by_example.ipynb
@@ -15,9 +15,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "\n",
-    "\n",
-    "There are lots of non-FastHTML-specific tricks and patterns involved in building web apps. The main goal of this tutorial is to give an alternate introduction to FastHTML. Building out example applications to show common patterns used. Also, illustrate some of the ways you can build on top of the FastHTML foundations to create your own custom web apps. A secondary goal is to have this be a useful document to add to the context of an LLM to turn it into a useful FastHTML assistant.\n",
+    "This tutorial provides an alternate introduction to FastHTML by building out example applications. We also illustrate how to use FastHTML foundations to create custom web apps. Finally, this document serves as minimal context for a LLM to turn it into a FastHTML assistant.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


jph00:
This is great! BTW the text "This is needed to tell FastHTML to turn the main body into a full HTML page that includes the headers (including the pico link and our custom css) which we passed in" is wrong. A single FT element is sufficient for this to work (cc @johnowhitaker since IIRC you wrote that bit)

jph00:
@pydanny you might be interested in this PR to see some nice nbdev/quarto features icymi

hamelsmu:
I'm a little confused by this

> "This is needed to tell FastHTML to turn the main body into a full HTML page that includes the headers (including the pico link and our custom css) which we passed in" is wrong. A single FT element is sufficient for this to work (cc @johnowhitaker since IIRC you wrote that bit)

I can't seem to figure out how to change the code exactly to do this without a tuple? 

```python
from fasthtml.common import *
css = Style(':root {--pico-font-size:90%,--pico-font-family: Pacifico, cursive;}') # <1>
app = FastHTML(hdrs=(picolink, css)) # <2>

@app.route("/")
def get():
    return (Title("Hello World"), 
               Main(H1('Hello, World'), cls="container"))
```

jph00:
I've updated that sentence now to be more accurate.

PR #411, opened 2024-09-09
Add note on SVG components imports

callmephilip:
**Proposed Changes**

Adds a note regarding explicit SVG element imports 

**Types of changes**

Documentation update

**Checklist**


jph00:
You don't have to explicitly import them -- a wildcard import works fine.

callmephilip:
> You don't have to explicitly import them -- a wildcard import from `fasthtml.svg` works fine too.

good point. i've updated the instructions

jph00:
Thank you.

PR #417, opened 2024-09-12
fix generation of `llms-ctx*.txt` files

hamelsmu:
The flags were wrong in the script!  Threw me off.  Also the published versions of these are wrong too.  

https://docs.fastht.ml/llms-ctx.txt 

vs

https://docs.fastht.ml/llms-ctx-full.txt

Are the same 😆 .  But no worries.  This fixes that.


hamelsmu:
Made this change in fastcore too with this commit https://github.com/fastai/fastcore/commit/3b9f27f462a71e65d06f8a1b92b5f15da00f499a

jph00:
ha! thank you :D

PR #421, opened 2024-09-13
Add documentation explainer for websockets with more advanced example

ncoop57:
---
name: WebSocket Docs
about: This PR adds more documentation around the usage of websockets
title: '[PR] '
labels: ''
assignees: ''

---

**Proposed Changes**
FastHTML supports websockets, however, the documentation for it is lacking and this PR attempts to address with by adding more clarity and examples on what can be done with websockets in FastHTML and a more advanced app than the basic one currently available

**Types of changes**

**Checklist**


algal:
It's great to see some more documentation for this area.

Here are some questions readers might have:

1. Why are the socket route handlers async, and what does that imply I should be careful of? (My own understanding is that they're async because they need to be in order to use send, but that this also implies they are suspendable in a way that could mutating across different requests if the functions relied any mutable state as opposed to being pure functions. But I'm not 100% sure of this!)
2. When you say "The client will then update the message div with the message from the server using Out of Band Swaps" what in the code makes that happen? I don't see a div that is obviously the message div, and I don't see code that seems to explicitly specify an OOB swap or any particular swap type.

examples/chat_ws.py
```diff
@@ -0,0 +1,32 @@
...
+def on_disconnect(ws):
+    connection_id = str(id(ws))
+    if connection_id in users:
+        users.pop(connection_id)
+
+@app.ws('/ws', conn=on_connect, disconn=on_disconnect)
+async def ws(msg:str):
+    messages.append(msg)
+    for u in users.values():
+        await u(Ul(*[Li(m) for m in messages], id=nid))
```
pydanny, line 30: Consider adding a comment that this is affecting the `users` dict otherwise some might interpret this as `await` returning a value much in the same way `return` and `yield` do.

pydanny (changes requested):
Fantastic! I really appreciate the minimalist code example.

I just have one request which is that there's ambiguouty that can trip people up. A code comment should address it. 🙂

examples/chat_ws.py
ncoop57: Added, lmk if that comment makes more sense now or if you have a suggetion for the wording lmk!

Isaac-Flath:
I've been working on integrating this example into the gallery.  Here are the changes that I think I am going to include might be relevant here as well:
- `str(id(ws)` - I don't think the `str` casting is needed
- made a `render_messages` function to replace the duplication `Ul(*[Li(m) for m in msgs], id='msg-list')`
- Used a `mk_input` function to both render the main page, but also reset the input in the web socket.  Described [here](https://discord.com/channels/1200111522916094103/1284220385956593775/1284611147101896705)
- Got rid of `id='form'` since the id isn't used anywhere.

Also some other changes for aesthetics, routing to make it work in the gallery that probably shouldn't be in this explainer example so I'm not listing those here.

audreyfeldroy:
Amazing @Isaac-Flath! I'm looking through the gallery and its repo...can I trouble you to post or DM me a link to what you've got so far? I think this PR will probably end up using most of what you have. 

I'm also wondering if for a minimal example it makes sense to keep this form a bit more vanilla, using a Send button as the submit, to give our audience an example that handles accessibility by default. (Unless FastHTML or HTMX have something that auto-adds ARIA roles and Enter key handling when an input acts as the whole form including submit - not sure?)

Isaac-Flath:
I just pushed what I have so far. It's not finalized, and I haven't deployed it, but the code is [here](https://github.com/AnswerDotAI/FastHTML-Gallery/blob/main/examples/dynamic_user_interface/web_sockets/app.py).

pydanny (approved):

Issue #432, opened 2024-09-16
[BUG] Fix Uncaught SyntaxError


pydanny:
**Describe the bug**

On JS load this error happens:

> Uncaught SyntaxError: The requested module 'https://cdn.jsdelivr.net/gh/answerdotai/fasthtml-js/fasthtml.js' does not provide an export named 'proc_htmx'

Amongst other things, this breaks markdown rendering. See photos at bottom for example.

**Minimal Reproducible Example**

```python
from fasthtml.common import *

app, rt = fast_app(hdrs=[MarkdownJS(), HighlightJS()])

@rt("/")
def home(): return Div("Hello, World")

serve()
```

**Expected behavior**

Calling proc_htmx shouldn't generate errors

**Environment Information**
Please provide the following version information:
- fastlite version: 0.0.11
- fastcore version: 1.7.8
- fasthtml version: 0.6.0

**Confirmation**
Please confirm the following:
- [x] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [x] I have provided a minimal reproducible example
- [x] I have included the versions of fastlite, fastcore, and fasthtml
- [ ] I understand that this is a volunteer open source project with no commercial support.

**Screenshots**

<img width="741" alt="Screenshot 2024-09-16 at 16 55 33" src="https://github.com/user-attachments/assets/953fcbce-9556-4e0c-ba53-720fed807ba8">

<img width="803" alt="Screenshot 2024-09-16 at 16 56 03" src="https://github.com/user-attachments/assets/78ad7e9b-420a-4eac-850a-6460d978dcf8">

jph00:
Fixed by #434

PR #433, opened 2024-09-16
Fix proc_htmx bug

pydanny:
---
name: Fix proc_htmx bug
about: Corrects the error: `Uncaught SyntaxError: The requested module 'https://cdn.jsdelivr.net/gh/answerdotai/fasthtml-js/fasthtml.js' does not provide an export named 'proc_htmx'`
title: '[PR] Fix proc_htmx bug'
labels: 'bug'
assignees: '@jph00'

---

**Related Issue**

#432 

**Proposed Changes**

Corrects the JS import to not point at just the `proc_htmx` function, which is already loaded.

**Types of changes**

**Checklist**

### DIFF
--- a/fasthtml/js.py
+++ b/fasthtml/js.py
@@ -27,7 +27,7 @@ def dark_media(
 
 # %% ../nbs/api/03_js.ipynb
 marked_imp = """import { marked } from "https://cdn.jsdelivr.net/npm/marked/lib/marked.esm.js";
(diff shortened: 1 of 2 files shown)


jph00:
We want to use this as ESM here. So importing it globally isn't actually the goal. I'll close this PR, since it's not the desired solution -- could you instead please create an MRE and open an issue showing the problem you're facing?

PR #440, opened 2024-09-19
Update PageX function

dgwyer:
---
name: Pull Request
title: '[PR] '
labels: ''
assignees: ''

---

**Proposed Changes**
This PR updates the `PageX` function to optionally accept a named `h1` argument that will override the main `H1` tag, so that it can be different to the browser title text.

With the updated component you can now do something like this to have different text for the `H1` tag:

```
PageX('Title', Ul(Li('One'), Li('Two'), Li('Three')), P('Just a paragraph'), h1='H1 Heading')
```

This will add 'Title' as the browser tab title, and 'H1 Heading' as the main page `H1` tag text.

Note: This does not break any current uses of `PageX` as the new named parameter is entirely optional. So this will still work as expected:

```
PageX('Title', Ul(Li('One'), Li('Two'), Li('Three')), P('Just a paragraph'))
```

This will add 'Title' as the browser tab title, **and** as the main page `H1` tag.

**Types of changes**

**Checklist**

**Additional Information**
My motivation for this came about when I was looking at doing a PR for the [fasthtml-tut](https://github.com/AnswerDotAI/fasthtml-tut) repo to update the examples there. I thought it would be nice to be able to use `PageX()` to add the Python file as the browser title, and the relevant text (e.g. 'Todo list') as the `H1` text.

At the moment you can't do this as the first positional argument for `PageX` is used for the browser title **and** `H1` tag.

### DIFF
--- a/fasthtml/fastapp.py
+++ b/fasthtml/fastapp.py
@@ -94,4 +94,7 @@ def fast_app(
 
 # %% ../nbs/api/10_fastapp.ipynb
 def ContainerX(*cs, **kwargs): return Main(*cs, **kwargs, cls='container', hx_push_url='true', hx_swap_oob='true', id='main')
-def PageX(title, h1=None, *con): return Title(title), ContainerX(H1(h1 if h1 else title), *con)
+def PageX(title, *con, **kwargs):
+    h1_title = kwargs['h1'] if 'h1' in kwargs else title
+    return Title(title), ContainerX(H1(h1_title), *con)
+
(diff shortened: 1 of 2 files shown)


dgwyer:
Thinking about this we could also consider instead overriding the browser title rather than H1 tag text. I guess it depends on which one you'd expect to want to override less, which is probably the browser title. I can easily switch it around so this can be overridden rather than the H1 tag text.

dgwyer:
I think the `PageX` component has been completely removed now in the latest release? If so we can delete this PR.

jph00:
I actually removed that function the other day, sorry!

PR #453, opened 2024-09-23
Improve type annotations for element attributes

callmephilip:
---
name: Improve type annotations for html/htmx attributes
title: '[PR] '
labels: ''
assignees: ''

---

**Proposed Changes**

There is a whole lot of attributes associated with DOM elements that are currently typed to `Any`. This leads to the situations like this

before

Developers need to jump to HTMX doc to figure out what the valid values for those attributes are. It would be great to get better type annotations like this

after

I've added a sample annotation for htmx attribute `hx_swap` to demonstrate the idea + approach. Furthermore, proper annotations can be introduced on the `ft_hx` level doing something like this:

```python
def annotate_args(annotations = {}):
  def _f(f):
    sig = inspect.signature(f)
    sigd = dict(sig.parameters)
    sig_update = {
       n: inspect.Parameter(
           n,
           inspect.Parameter.KEYWORD_ONLY,
           default=None,
           annotation=t
          ) for n, t in annotations.items() 
    }
    sigd.update(sig_update)
    f.__signature__ = sig.replace(parameters=sigd.values())
    return f
  return _f

@annotate_args(annotations=hx_attrs_annotations)
@use_kwargs(hx_attrs, keep=True)
def ft_hx(tag: str, *c, target_id=None, hx_vals=None, **kwargs):
    ...
``` 

Or alternatively propagating this to `@use_kwargs` to add support for annotations:

```python
def mk_param(n, d=None, annotation=inspect.Parameter.empty): 
    return inspect.Parameter(n, inspect.Parameter.KEYWORD_ONLY,  default=d, annotation=annotation)

def use_kwargs(names, keep=False, annotations={}):
    "Decorator: replace `**kwargs` in signature with `names` params"
    def _f(f):
        sig = inspect.signature(f)
        sigd = dict(sig.parameters)
        k = sigd.pop('kwargs')
        s2 = {n:mk_param(n, annotation=annotations.get(n, inspect.Parameter.empty)) for n in names if n not in sigd}
        sigd.update(s2)
        if keep: sigd['kwargs'] = k
        f.__signature__ = sig.replace(parameters=sigd.values())
        return f
    return _f
```

**Types of changes**

**Additional Information**
This is really an initial proposal. Obviously more work is required for this to be useful


jph00:
Sure I'm opening to considering this interesting idea. This PR needs a little bit of cleanup. There's an extra cell that you've accidentally left in with output from a `?` command, also, your list of literal options are all on separate lines. We really like to avoid using vertical space wherever possible, so please put these together. 

Could you add the other HTMX annotations into this PR as well please? I want to see how many there are and how big they are, so I can check whether I'm happy to put them directly in the notebook.

callmephilip:
@jph00 this should be ready for review after https://github.com/AnswerDotAI/fasthtml/pull/457

callmephilip:
now that i think about it, does `hx_attrs_annotations` sound too mouthful? should this be `hx_attrs_types`?

jph00:
This PR looks pretty good actually - I'll merge it, but let's see what the feedback is. I wonder if people will find the choice popups confusing, since other strings are OK as well (as you correctly identified with the `| str`). But yes let's use `hx_attrs_types` -- send a followup PR for that please.

PR #454, opened 2024-09-23
Allow custom messages to be outputted when the terminal starts

dgwyer:
---
name: Pull Request
title: '[PR] '
labels: ''
assignees: ''

---

**Proposed Changes**
This PR allows users to specify additional terminal messages, outputted when the server starts. By default a terminal message linking to the localhost web app is outputted:

```
Link: http://localhost:5001
```

I thought that it would be useful to be able to optionally specify custom messages too. For example, when your app includes a db then it is useful in development to be able to view/edit the db in the browser. If I wrap the `serve()` function and run the db viewer in a subprocess a minimal app would look something like this:

```
from fasthtml.common import * 
import subprocess

app,rt,todos,Todo = fast_app('data/todos.db', id=int, title=str, done=bool, pk='id')

@rt("/")
def get():
	return Div('Welcome to FastHTML!')

def serve_db(db_path='data/data.db', port=8083):
    sqlite_process = subprocess.Popen(
        ['sqlite_web', db_path, '--port', str(port), '--no-browser'],
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL
    )

    try:
        serve(msg=(f'SQLite: http://localhost:{port}'))
    finally:
        sqlite_process.terminate()

serve_db()
```

Notice the call to `serve()` passes in a `msg` argument:

`serve(msg=(f'SQLite: http://localhost:{port}'))`

This PR updates the `serve()` function to accept a string, or tuple of string messages to be outputted after the default meassage `Link: http://localhost:5001`, and produces output like this:

image

Notice that below the default message, there is now an additional message that outputs a link that opens up the db viewer/editor in the browser.

image

**So, now we have a way to conveniently add to the default terminal message for inserting links to other running services, or anything else you like!**

**Types of changes**

**Checklist**

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -588,7 +588,8 @@ def serve(
         port=None, # If port is None it will default to 5001 or the PORT environment variable
         reload=True, # Default is to reload the app upon code changes
         reload_includes:list[str]|str|None=None, # Additional files to watch for changes
-        reload_excludes:list[str]|str|None=None # Files to ignore for changes
+        reload_excludes:list[str]|str|None=None, # Files to ignore for changes
+        msg:list[str]|None=None, # List of terminal messages to print
         ): 
     "Run the app in an async server, with live reload set as the default."
     bk = inspect.currentframe().f_back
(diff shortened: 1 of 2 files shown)


jph00:
I think I'll just let people use print(), even if it's a bit of extra noice!

PR #455, opened 2024-09-23
Added missing IPython dependency

mone27:
After the update in 0.6.5 that adds jupyter support IPython is required, but as it was missing from the dependencies is not installed by pip/conda

**Types of changes**


jph00:
I'd rather not make this a dep. Let's find another way to fix it.

PR #461, opened 2024-09-24
Add notes re: contributing to FastHTML

callmephilip:
---
name: Add notes re: contributing to FastHTML
about: Contribution docs update
title: '[PR] '
labels: ''
assignees: ''

---

A couple of notes based on my experience contributing to the codebase


jph00:
We should add `nbdev_clean` as a thing to run before committing too.

callmephilip:
> We should add `nbdev_clean` as a thing to run before committing too.

how about setting up githooks for these things maybe?

jph00:
We do have `nbdev_install_hooks`, although for a quick "getting started" that can feel like extra work.

PR #472, opened 2024-09-25
[OAuth] remove default scope for GitHub

hamelsmu:
It works fine without any scope.  I think its unexpected that it tries to set a user scope for you


jph00:
Thanks!

PR #475, opened 2024-09-26
add missing Leave Room button in e2e tutorial

ARK3r:
Missing leave room button in step by step setup of the room page.


jph00:
@ncoop57 want to take a look?

ncoop57:
Ah good catch @ARK3r Thanks for the fix!

Issue #483, opened 2024-09-30
[BUG] OAuth URI Errors


DancingPete:
# Bug Description

According to your [OAuth documentation](https://docs.fastht.ml/explains/oauth.html#a-work-in-progress), I should be able to call client.login_link() to generate the href for the Oauth redirect handler. However, the sever errors out with the following message:

`TypeError: WebApplicationClient.login_link() missing 1 required positional argument: 'redirect_uri'`

# Reproducible Example

```python
# Standard library imports
from os import environ

# Third-party imports
from fasthtml.common import *
from fasthtml.oauth import *

# Constants
GOOGLE_OAUTH_CLIENT_ID = environ.get("GOOGLE_OAUTH_CLIENT_ID")
GOOGLE_OAUTH_CLIENT_SECRET = environ.get("GOOGLE_OAUTH_CLIENT_SECRET")
GOOGLE_OAUTH_REDIRECT = environ.get("GOOGLE_OAUTH_REDIRECT")

# App definition
app, rt = fast_app(debug=environ.get("DEBUG"), live=environ.get("LIVE"))

# Define Google app client
client = GoogleAppClient(
    client_id=GOOGLE_OAUTH_CLIENT_ID,
    client_secret=GOOGLE_OAUTH_CLIENT_SECRET,
    redirect_uri=GOOGLE_OAUTH_REDIRECT)
link = client.login_link()

@rt("/")
def get(): return A(href=link)


@rt("/redirect")
def get(code: str): return P(f"code: {code}")
```
It's worth noting that in creating a breakpoint before calling the client.login_link and examining the attributes of client, there is no redirect_uri, but instead a redirect_url variable.

Also, in supplying the the redirect uri directly in the login_link() function as an argument, the link is generate, but the client fails at the Google authentication screen (which may or may not be related).

# Expected Behaviour
I would expect the client.login_link() to produce the redirect necessary to the oauth flow.

# Environment Information
Please provide the following version information:
- fastlite version: 0.0.11
- fastcore version: 1.7.10
- fasthtml version: 0.6.9

# Confirmation
Please confirm the following:
- [x] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [x] I have provided a minimal reproducible example
- [x] I have included the versions of fastlite, fastcore, and fasthtml
- [x] I understand that this is a volunteer open source project with no commercial support.

# Additional context

# Screenshots

algal:
[...]

PR #491, opened 2024-10-03
Update OAuth explainer notebook

algal:
Updated it not to hardcode the auth callback URL but to drive the URL from the path at all reference sites in the code.

Updated the notebook to explain the flow more fully.

Tested and working locally.

---
name: Pull Request
title: '[PR] Update OAuth explainer notebook'
labels: ''
assignees: ''

---

**Related Issue**
This was originally associated with the following issue where the code shown in docs did not work: https://github.com/AnswerDotAI/fasthtml/issues/483

This PR is a follow-up fix to avoid hardcoding the authorization callback URL and to improve explanation, for those who do not already know oauth.

**Proposed Changes**
Removes hardcoded authorization callback URL from the sample code in the documentation, and improve the explanation in the documentation.

I have used a slightly more verbose style for the sample code than is typical of project code, because its main function is to teach oauth.

**Types of changes**
- [X] Bug fix (non-breaking change which fixes an issue)

**Checklist**
- [X] My code follows the code style of this project.
- [X] I have updated the documentation accordingly.
- [X] I am aware that this is an nbdev project, and I have edited, cleaned, and synced the source notebooks instead of editing .py or .md files directly.

**Additional Information**


jph00:
Great! Can you please remove the big code listing at end of this page and instead encourage people to read the fasthtml-example code? Also, could we add docs around using the `oa.py` class-based approach to this NB? (I'll merge this, so these can be separate PRs)

PR #506, opened 2024-10-09
Update live reload documentation

siralmat:
The Hello World example at the beginning of the [tutorial](https://docs.fastht.ml/tutorials/by_example.html) returns a raw HTML string, so it's not compatible with live reload. For the hypothetical eager reader who follows the callout tip and enables live reloading before continuing with the tutorial, this is tricky to debug. 

Changes:
- Add a note to the Live Reload "Gotchas" section about the need to use ft components 
  - Is this the correct term? The docs seem to use "ft components", "FT components", and "FT objects" interchangeably.
- Move the live reload callout tip further down in the tutorial after components have been introduced
  - I didn't want to make any major changes to the tutorial flow, but personally I think it would be a nicer learning experience to just dive straight into components instead of starting with an example that's not _really_ a valid FastHTML app.
- Replace deprecated `run_uv()` reference in `FastHTMLWithLiveReload`

Thanks for the awesome library, looking forward to reading the rest of the tutorial! 😁

### DIFF
--- a/fasthtml/live_reload.py
+++ b/fasthtml/live_reload.py
@@ -47,7 +47,7 @@ class FastHTMLWithLiveReload(FastHTML):
         >>> app = FastHTMLWithLiveReload()
 
         Run:
-            run_uv()
+            serve()
     """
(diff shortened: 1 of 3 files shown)


jph00:
Great points - thanks!

Issue #513, opened 2024-10-11
[BUG] Uploading a single file on a multiple file field requires try/except


pydanny:
**Describe the bug**

If a file input is set to `multiple=True` and only a single file is selected on upload the form handler requires this try/except block or it throws a `TypeError`:

```python
@rt('/upload-multiple')
async def post(files: list[UploadFile]):
    try: iter(files)
    except TypeError: files = [files]
```

**Minimal Reproducible Example**

```python
from fasthtml.common import *
from pathlib import Path

app, rt = fast_app()

upload_dir = Path("filez")
upload_dir.mkdir(exist_ok=True)

@rt('/')
def get():
    return Titled("File Upload Demo",
        P('Upload 1 to display failure but multiples work fine'),
        Article(
            Form(Input(type="file", name="files", multiple=True),
                Button("Upload", type="submit", cls='contrast'),
                hx_post="/upload-multiple", hx_target="#result-multiple"),
            Div(id="result-multiple")
        )            
    )

def FileMetaDataCard(file):
    return Article(
        Header(H3(file.filename)),
        Ul(Li('Size: ',file.size),Li('Content Type: ',file.content_type))
    )    

@rt('/upload-multiple')
async def post(files: list[UploadFile]):
    # Uncomment to show current method to resolve the problem described
    # try: iter(files)
    # except TypeError: files = [files]
    cards = []
    for file in files:
        filebuffer = await file.read()
        (upload_dir / file.filename).write_bytes(filebuffer)
        cards.append(FileMetaDataCard(file))
    return cards    

serve()
```

**Expected behavior**

The `list[UploadFile]` type on a handler's arguments should cast a single submitted `UploadFile` to `list[UploadFile]`. Something like:

```python
try: iter(files)
except TypeError: files = [files]
```

**Environment Information**
Please provide the following version information:
- fastlite version: 0.0.11
- fastcore version: 1.7.13
- fasthtml version: 0.6.10

**Confirmation**
Please confirm the following:
- [x] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [x] I have provided a minimal reproducible example
- [x] I have included the versions of fastlite, fastcore, and fasthtml
- [x] I understand that this is a volunteer open source project with no commercial support.

PR #514, opened 2024-10-11
Fix single file uploads in multi-file forms

pydanny:
---
name: Pull Request
title: '[PR] Fix single file uploads in multi-file forms'
labels: 'bug'
assignees: ''

---

**Related Issue**

#513 Uploading a single file on a multiple file field requires a try/except block

**Proposed Changes**

If the annotation for the response element says it is a `GenericAlias` (which roughly means 'iterable') but the response element isn't actually iterable, then wrap it in a list.

**Types of changes**

**Checklist**

**Additional Information**

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -204,6 +204,8 @@ async def _find_p(req, arg:str, p:Parameter):
     except ValueError: raise HTTPException(404, req.url.path) from None
 
 async def _wrap_req(req, params):
+    # If [UF, UF] then it behaves on way
+    # If [UF] then it behaves another way
     return [await _find_p(req, arg, p) for arg,p in params.items()]
 
 # %% ../nbs/api/00_core.ipynb
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -193,6 +193,10 @@ async def _find_p(req, arg:str, p:Parameter):
     if (res in (empty,None)) and p.default is empty: raise HTTPException(400, f"Missing required field: {arg}")
     # If we have a default, return that if we have no value
     if res in (empty,None): res = p.default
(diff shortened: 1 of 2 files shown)


pydanny:
Hey @jph00, here's a patch for multi-file uploads that comes with a test.

fasthtml/core.py
```diff
@@ -191,6 +191,10 @@ async def _find_p(req, arg:str, p:Parameter):
     if (res in (empty,None)) and p.default is empty: raise HTTPException(400, f"Missing required field: {arg}")
     # If we have a default, return that if we have no value
     if res in (empty,None): res = p.default
+    # if anno is a Generic Alias but res isn't an iterable then wrap in list
+    if isinstance(anno, GenericAlias):
```
jph00, line 195: I think there are parameterised generics that aren't lists? We should probably me more specific about what we're looking for.

jph00 (changes requested):
Thanks Danny! Just one suggestion to look at.

fasthtml/core.py
pydanny: @jph00 The challenge is that for this error `anno` is a `GenericAlias`, which arguably is a bug in its own right. Or maybe we're just not particular enough so this is a missing feature? Looking into that now.

In the meantime, I've changed the `try/except` to a more precise check. Hopefully that goes away, replaced with a more accurate `anno` specification.

pydanny: Diving into Fastcore's [signature_ex](https://fastcore.fast.ai/basics.html#signature_ex) function.

pydanny: Hmmm... I think `signature_ex` is a tangent. Digging into the behavior around params in `_wrap_req`.

PR #563, opened 2024-11-04
Incl session param in ws

callmephilip:
---
name: Pull Request
title: '[PR] '
labels: ''
assignees: ''

---

**Related Issue**
In order to access session data in ws handler right now, you need to drill into `ws.session`. While standard route handlers provide `session/sess` shortcut, `ws` handlers do not.

**Proposed Changes**
This PR adds support for `session/sess` param inside `ws` handlers. Includes a quick test + doc 

**Types of changes**

**Checklist**


jph00:
Good idea! :D

Issue #637, opened 2025-01-19
Handling for query or url params in qp/to


Isaac-Flath:
Allow qp (and therefore `.to` method) to handle params with `?a=1` or `/a` situations

PR #638, opened 2025-01-19
Handling for query or url params in qp/to

Isaac-Flath:
---
name: Pull Request
about: Allow qp (and therefore `.to` method) to handle params with `?a=1` or `/a` situations
title: '[qp improvement] '
labels: 'enhancement'
assignees: 'isaac-flath'

---

**Related Issue**
[Please link to the issue that this pull request addresses. If there isn't one, please create an issue first.
](https://github.com/AnswerDotAI/fasthtml/issues/637)

**Proposed Changes**
Allow qp (and therefore `.to` method) to handle params with `?a=1` or `/a` situations

**Types of changes**

**Checklist**

**Additional Information**

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -477,17 +477,16 @@ def _wrap_ex(f, status_code, hdrs, ftrs, htmlkw, bodykw, body_wrap):
 
 # %% ../nbs/api/00_core.ipynb
 def qp(p:str, **kw) -> str:
-    "Add query parameters to path p"
+    "Add parameters kw to path p"
     kw = {k.split(':')[0]:v for k,v in kw.items()}
-    for k,v in kw.copy().items(): # Go through all params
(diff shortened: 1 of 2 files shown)


jph00:
Will this break for path params with a `:` in them?

Isaac-Flath:
@jph00 I've added a test for that now; please double check that's the behavior we should be expecting!

jph00:
I would expect to be using `ab` as the key - the bit after the `:` isn't really part of the name IMO; wdyt?

jph00:
@Isaac-Flath just making sure you saw my comment above.

Isaac-Flath:
@jph00  I think this looks good - What do you think?

jph00:
It's not perfect, but it's a fine start so I'll merge and we can discuss improvements.

Issue #641, opened 2025-01-28
[FEATURE] Reconnect live reload WS after server error


dgwyer:
At the moment when using the live reload feature, if you accidentally make a typo then the web socket is closed and the browser displays an "Internal Server Error" message. Even once you correct the typo and the server restarts, the WS connection does not automatically reconnect until you manually refresh the browser.

This is a little annoying and would be nice if the WS reconnects automatically after a restart from a server error.

**Confirmation**
Please confirm the following:
- [X] I have checked the existing issues and pull requests to ensure this feature hasn't been requested before.
- [X] I have read the project's documentation to ensure this feature doesn't already exist.

dgwyer:
Closed as completed.

PR #642, opened 2025-01-28
Reload FastHTML app after server error

dgwyer:
**Proposed Changes**
For a FastHTML app that has live reload enabled, after a server error occurs (e.g. triggered by a typo in your app source code) the web socket does not automatically reconnect. This PR fixes that. When the server restarts from an error the WS is automatically reconnected.

See this [related issue](https://github.com/AnswerDotAI/fasthtml/issues/641) for more details.

**Types of changes**
- [X] Bug fix (non-breaking change which fixes an issue)

**Checklist**
- [X] My code follows the code style of this project.
- [X] I am aware that this is an nbdev project, and I have edited, cleaned, and synced the source notebooks instead of editing .py or .md files directly.

**Additional Information**

### DIFF
--- a/fasthtml/live_reload.py
+++ b/fasthtml/live_reload.py
@@ -3,24 +3,42 @@ from fasthtml.basics import FastHTML, Script
 
 __all__ = ["FastHTMLWithLiveReload"]
 
-
-def LiveReloadJs(reload_attempts:int=1, reload_interval:float=1000., **kwargs):
+def LiveReloadJs(reload_attempts:int=10, reload_interval:float=1000., server_check_attempts:int=20, **kwargs):
     src = """
-    (function() {
-        var socket = new WebSocket(`ws://${window.location.host}/live-reload`);
-        var maxReloadAttempts = %s;
-        var reloadInterval = %s; // time between reload attempts in ms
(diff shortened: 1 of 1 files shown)


curtis-allan:
Thanks for taking the time to craft this PR @dgwyer. Definitely agree that this would be a nice addition to local development.

A couple of notes:
- Clever use of the fetch request to check for page status vs the ws route!
- Noticed you changed reload_attempts but it's not being passed to the formatted string, was there a reason for this?
- The changed code is quite large for what it does

You have some great functionality here, tested it out locally and it works well. I might close this PR and make a new one with some tweaks while keeping it functionally similar.

Thanks again for the work on this enhancement, we will credit your efforts in the release notes :)

dgwyer:
@curtis-allan Thanks for the feedback!

> Noticed you changed reload_attempts but it's not being passed to the formatted string, was there a reason for this?

I'm not sure if I remember now. 🙂 Maybe I tried something that I didn't follow-up with.

I also tried to reduce the reload_interval in a separate test (which I didn't submit a PR for). I really wanted to improve the reload speed to improve the UX when live reload is enabled. Perhaps you could also look into this in your PR?

PR #650, opened 2025-02-07
Make the H1 tag optional, so you can add your own

dgwyer:
---
name: Pull Request
about: Make the `Titled()` a little more flexible by making the default generated H1 tag optional
title: '[PR] Make the H1 tag optional, so you can add your own'
labels: ''
assignees: ''
---

**Proposed Changes**
`Titled()` is a useful component but I don't sometimes use it as I need to specify a different `H1` text for the page heading. Or perhaps I need to add it inside a navbar rather than it be the first direct child of `Main()`.

e.g.
```python
from fasthtml.common import *

app,rt = fast_app(live=True)

@rt('/')
def get(): return Titled(
	"Page Title",
	H1("My Page Title"),
	Div(P("Hello, world!!")),
	titled_h1=False
)

serve()
```

Or:
```python
from fasthtml.common import *

app,rt = fast_app(live=True)

@rt('/')
def get(): return Titled(
	"Page Title",
	Div(
		H1("My Page Title"),
		P("Hello, world!!")
	),
	titled_h1=False
)

serve()
```

**Types of changes**

**Checklist**

### DIFF
--- a/fasthtml/xtend.py
+++ b/fasthtml/xtend.py
@@ -180,9 +180,11 @@ def jsd(org, repo, root, path, prov='gh', typ='script', ver=None, esm=False, **k
 
 # %% ../nbs/api/02_xtend.ipynb
 @delegates(ft_hx, keep=True)
-def Titled(title:str="FastHTML app", *args, cls="container", **kwargs)->FT:
+def Titled(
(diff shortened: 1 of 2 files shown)


dgwyer:
Forgot I submitted a similar PR a little while back. :)

Issue #635, opened 2025-01-12
[BUG] Toast does not works when session is typed


barenko:
The toast does not shows if you use typing in your session parameter.

Look the code:

```python

app,rt = fast_app()

setup_toasts(app)

@rt('/test-toast1')
def get(session): 
    add_toast(session, f"This is a test toast", "info")
    return Main(Div(f"I like toast, the session type is: {type(session)}"))

@rt('/test-toast2')
def get(session:dict): 
    add_toast(session, f"This is a test toast", "info")
    return Main(Div(f"I like toast, the session type is: {type(session)}"))
```

When I call /test-toast1, the toast shows and the msg says that session type is dict
When I call /test-toast2, the toast DOES NOT shows and the msg says that session type is dict

**Expected behavior**
The toast should work independently of the typing mark.

**Environment Information**
- fastlite version: 0.1.1
- fastcore version: 1.7.28
- fasthtml version: 0.10.3

**Confirmation**
Please confirm the following:
- [x] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [x] I have provided a minimal reproducible example
- [x] I have included the versions of fastlite, fastcore, and fasthtml
- [x] I understand that this is a volunteer open source project with no commercial support.

PR #651, opened 2025-02-07
handle for when there is a type hint on session

henriwoodcock:
---
name: Pull Request
title: '[PR] '
labels: ''
assignees: ''

---

**Related Issue**
https://github.com/AnswerDotAI/fasthtml/issues/635

**Proposed Changes**
I saw the issue and that when typed hinted, session was being handled as body. I was not sure whether to create or new function or just handle before as body is also a `dict`. Looks like this may also be true for other args like `auth` which I can add here (but wasn't sure it was ever a function arg).

**Types of changes**

**Checklist**
- [ ] I am aware that this is an nbdev project, and I have edited, cleaned, and synced the source notebooks instead of editing .py or .md files directly.

**Additional Information**


henriwoodcock:
force push to remove accidentally commited change

jph00:
Thanks!

Issue #336, opened 2024-08-19
[FEATURE] Starlette background tasks with FastHTML


pydanny:
**Is your feature request related to a problem? Please describe.**

It would be nice to have easy, elegant Background Tasks in FastHTML.  While FastHTML is async first and there's lots of ways to use that, having a standard way to handle Background Tasks would help a lot of people.

**Describe the solution you'd like**

Simple interface that behaves close to any other FastHTML response.

**Example code**
Provide an example of how you imagine the feature working:

```python
from fasthtml.common import *
from starlette.background import BackgroundTask
from asyncio import sleep
from time import time

app, rt = fast_app()

async def counter(loops):
    for i in range(loops):
        print(i)
        await sleep(i)

@rt
async def index():
    task = BackgroundTask(counter, loops=5)
    return Response(to_xml(Titled(str(time()))), background=task)

serve()
```

Currently this returns:

```html
<title>1741832742.041034</title>
<main class="container">  <h1>1741832742.041034</h1>
</main>
```

Solutions:

1. Change the `FtResponse` type to handle Background tasks. But how to make this work with the current design? Am I misunderstanding the `FtResponse` class?
2. Add a dedicated `FtBackgroundResponse` type
3. Make the background task an argument to the route handler in the same style as FastAPI. See https://fastapi.tiangolo.com/tutorial/background-tasks/

**Problem solved**

Easy background tasks for the 95% use case.

**Confirmation**
Please confirm the following:
- [x] I have checked the existing issues and pull requests to ensure this feature hasn't been requested before.
- [x] I have read the project's documentation to ensure this feature doesn't already exist.

PR #674, opened 2025-03-13
Fix+docs for background tasks when used by FtResponse

pydanny:
---
name: Pull Request
title: '[PR] Correct background tasks when used by FtResponse'
labels: 'bug'
assignees: ''

---

**Related Issue**

#336 - Background Tasks

**Proposed Changes**

This PR makes it possible for the FtResponse class to handle Background Tasks

**Types of changes**

**Checklist**

**Additional Information**

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -781,7 +781,7 @@ class FtResponse:
 
     def __response__(self, req):
         cts,httphdrs,tasks = _xt_cts(req, self.content)
-        tasks = getattr(self, 'background', tasks)
+        if not tasks.tasks: tasks = self.background
         headers = {**(self.headers or {}), **httphdrs}
         return self.cls(cts, status_code=self.status_code, headers=headers, media_type=self.media_type, background=tasks)
 
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -775,12 +775,13 @@ class MiddlewareBase:
 # %% ../nbs/api/00_core.ipynb
 class FtResponse:
     "Wrap an FT response with any Starlette `Response`"
-    def __init__(self, content, status_code:int=200, headers=None, cls=HTMLResponse, media_type:str|None=None):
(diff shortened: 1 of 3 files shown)


fasthtml/core.py
```diff
@@ -773,12 +773,13 @@ async def __call__(self, scope, receive, send) -> None:
...
     "Wrap an FT response with any Starlette `Response`"
-    def __init__(self, content, status_code:int=200, headers=None, cls=HTMLResponse, media_type:str|None=None):
+    def __init__(self, content, status_code:int=200, headers=None, cls=HTMLResponse, media_type:str|None=None, background: BackgroundTask | None = None):
         self.content,self.status_code,self.headers = content,status_code,headers
-        self.cls,self.media_type = cls,media_type
+        self.cls,self.media_type,self.background = cls,media_type,background
 
     def __response__(self, req):
         cts,httphdrs,tasks = _xt_cts(req, self.content)
+        tasks = getattr(self, 'background', tasks)
```
jph00, line 782: I think `self.background` will *always* be set, so I'm not sure this is doing anything? Also IIUC it's the wrong way around - we should use the manually added `tasks` in the response if there, and only then fall back to `self.background`, shouldn't we?

Either way, it's a good idea to have tests or examples that exercise your code so that you know for sure whether it's working.

pydanny: Thanks for the feedback, we'll adjust the API accordingly and see about improving/adding to the tests and docs.

jph00:
Let me know when it's ready for review @pydanny (no hurry though)

pydanny:
This is ready for review.

hamelsmu:
@pydanny This is a really well written tutorial.  I love how you use code annotations for the docs with Quarto- extra love.  Made it really easy to read and also learn about how it works.

hamelsmu (approved):

hamelsmu:
@jph00 do you want to do a review or is this ok to merge?  I find this feature to be very valuable in most things that I end up building.

hamelsmu:
Edit:

Q: Can we ship this now?
A: Yes.

Merging

PR #680, opened 2025-03-17
Improve FastHTML concise guide

pydanny:
---
name: Pull Request
title: '[PR] Improve FastHTML concise guide'
labels: 'documentation'
assignees: audreyfeldroy, pydanny

---

**Proposed Changes**

Documentation enhancement

**Types of changes**

**Checklist**

**Additional Information**


drafts/concise_guide.ipynb
```diff
@@ -92,7 +92,66 @@
...
-    "When creating a response, FastHTML will automatically render tags using the to_xml function. "
+    "This example demonstrates key aspects of how FastHTML components handle attributes:\n",
+    "\n",
+    "```python\n",
+    "Label(\n",
+    "    \"Choose an option\", \n",
+    "    Select(\n",
+    "        Option(\"one\", value=\"1\", selected=True),  # Boolean True renders just the attribute name\n",
+    "        Option(\"two\", value=\"2\", selected=False), # Boolean False omits the attribute entirely\n",
+    "        Option(\"three\", value=3),                 # Non-string values are converted to strings\n",
```
jph00, line 103: Put kw args on a single line.

drafts/concise_guide.ipynb
```diff
@@ -92,7 +92,66 @@
...
+    "        cls=\"selector\",                           # 'cls' becomes 'class' in HTML (Python reserved word)\n",
```
jph00, line 104: We don't normally use `_id`, but just `id`, and we want to only show one way to do things. You can mention the underscore strip behavior in the markdown below briefly instead.

drafts/concise_guide.ipynb
```diff
@@ -92,7 +92,66 @@
...
    "metadata": {},
    "source": [
-    "When creating a response, FastHTML will automatically render tags using the to_xml function. "
+    "This example demonstrates key aspects of how FastHTML components handle attributes:\n",
+    "\n",
+    "```python\n",
+    "Label(\n",
+    "    \"Choose an option\", \n",
+    "    Select(\n",
+    "        Option(\"one\", value=\"1\", selected=True),  # Boolean True renders just the attribute name\n",
```
jph00, line 101: If you used `value=2` here, you could remove the next line and merge its comment into this one. i.e. Show as many things per line as possible!

drafts/concise_guide.ipynb
```diff
@@ -92,7 +92,66 @@
...
+    "        Option(\"three\", value=3),                 # Non-string values are converted to strings\n",
+    "        cls=\"selector\",                           # 'cls' becomes 'class' in HTML (Python reserved word)\n",
+    "        _id=\"counter\",                            # Leading underscore is stripped during rendering\n",
+    "        **{'@click':\"alert('Clicked');\"},         # Dict unpacking for attributes with special characters\n",
+    "    ),\n",
+    "    _for=\"counter\",                               # '_for' becomes 'for' (can also use 'fr')\n",
+    ")\n",
+    "```\n",
+    "\n",
+    "FastHTML components can be linked by matching `_id` with `_for` attributes. The library handles Python reserved words by using alternatives like `cls` for `class` and stripping leading underscores during rendering."
```
jph00, line 112: `id` and `for` matching is html behavior so doesn't need to be mentioned. The attr aliases were already mentioned earlier -- you could add the extra details to that part instead.

drafts/concise_guide.ipynb
```diff
@@ -92,7 +92,66 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can create new FT components in two ways, the preferred approach is to import the new component from `fasthtml.component`. If the component doesn't exist within that module, FastHTML will create it."
```
jph00, line 119: No need to mention two ways if we can get away with one! :) Also the comment and the markdown overlap a bit - try to remove the comment so info is in one place.

drafts/concise_guide.ipynb
```diff
@@ -161,6 +220,22 @@
...
+   "metadata": {},
+   "source": [
+    "Enums can be used to constrain the values accepted for a route parameter.\n",
+    "\n",
+    "```python\n",
+    "ModelName = str_enum('ModelName', \"alexnet\", \"resnet\", \"lenet\")\n",
+    "\n",
+    "@rt(\"/models/{nm}\")\n",
+    "def get(nm:ModelName): return nm\n",
+    "\n",
```
jph00, line 234: You can remove this since you're not actually printing it, and the output isn't important here.

drafts/concise_guide.ipynb
```diff
@@ -161,6 +220,22 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Enums can be used to constrain the values accepted for a route parameter.\n",
+    "\n",
+    "```python\n",
+    "ModelName = str_enum('ModelName', \"alexnet\", \"resnet\", \"lenet\")\n",
+    "\n",
```
jph00, line 231: We said that query params are idiomatic so let's use them.

drafts/concise_guide.ipynb
```diff
@@ -161,6 +220,22 @@
...
+    "@rt(\"/models/{nm}\")\n",
```
jph00, line 232: Perhaps we could combine the enum example into another example to show two things at once?

drafts/concise_guide.ipynb
```diff
@@ -468,6 +564,24 @@
     "print(client.get('/getcook').text)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "FastHTML provide access to Starlette's request object automatically using special `request` parameter name (or any prefix of that name)."
```
jph00, line 571: Put def body on previous line.

drafts/concise_guide.ipynb
```diff
@@ -488,6 +602,86 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "@rt\n",
+    "def htmlredirect(foo:FastHTML): return HtmxResponseHeaders(location=\"http://example.org\")"
```
jph00, line 619: You're not using `foo` here or referring to it, so either remove it or use it. Normally we'd just use the param `app` with no type IIRC so try that instead. Also see if you can merge this example with another one, especially since then it's clear you can return a tuple with headers along with fasttags etc.

drafts/concise_guide.ipynb
```diff
@@ -488,6 +602,86 @@
...
+   "source": [
+    "# products.py\n",
+    "ar = APIRouter(prefix=\"/products\")\n",
+    "\n",
+    "@ar(\"/all\")\n",
+    "def all_products(req):\n",
+    "    return Div(\n",
+    "        \"Welcome to the Products Page! Click the button below to look at the details for product 42\",\n",
+    "        Div(\n",
+    "            Button(\n",
```
jph00, line 643: Button can go on one line.

drafts/concise_guide.ipynb
```diff
@@ -488,6 +602,86 @@
...
+    "                hx_target=\"#products_list\",\n",
+    "                hx_swap=\"outerHTML\",\n",
+    "            ),\n",
+    "        ),\n",
+    "        id=\"products_list\",\n",
+    "    )\n",
+    "\n",
+    "\n",
+    "@ar.get(\"/{pid}\", name=\"details\")\n",
+    "def details(pid: int):\n",
```
jph00, line 655: One line here

drafts/concise_guide.ipynb
```diff
@@ -488,6 +602,86 @@
...
+    "                hx_get=req.url_for(\"details\", pid=42),\n",
+    "                hx_target=\"#products_list\",\n",
+    "                hx_swap=\"outerHTML\",\n",
+    "            ),\n",
+    "        ),\n",
+    "        id=\"products_list\",\n",
+    "    )\n",
+    "\n",
+    "\n",
+    "@ar.get(\"/{pid}\", name=\"details\")\n",
```
jph00, line 654: Use query params

drafts/concise_guide.ipynb
```diff
@@ -488,6 +602,86 @@
...
+    "def details(pid: int):\n",
+    "    return f\"Here are the product details for ID: {pid}\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Accessing the handler functions in the `APIRouter` object is done from the `rt_funcs` property"
```
jph00, line 663: That's not really relevant. Instead show how to import the routes into `app`.

jph00 (changes requested):
Looking good! I've made some fussy comments to help move in an even more concise direction... :)

drafts/concise_guide.ipynb
pydanny: We're confused, by query parameters do you mean the key/value that come after the question mark?

`/models?nm=alexnet`

Or do you mean this kind of parameter?

`/models/alexnet`

pydanny: We figured it out. We demonstrate query parameters later.

jph00:
You'll see in the guide I have the `func.to(key=val)` syntax mentioned,
which is the idiomatic way to use query params in fasthtml.

drafts/concise_guide.ipynb
```diff
@@ -670,25 +877,666 @@
...
+       "[User(id=3, name='Alma', active=0)]"
+      ]
+     },
+     "execution_count": null,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "users(\"name=?\", where_args=('Alma',))"
```
pydanny, line 1087: This added because otherwise AI often hallucinates a variation that doesn't work: `users(where="email=?", params=(email,))`

PR #692, opened 2025-03-28
Redux background tasks fix+docs

pydanny:
---
name: Pull Request
title: '[PR] fix+docs background tasks when used by FtResponse'
labels: 'bug'
assignees: 'pydanny,audreyfeldroy'

---

This PR combines PRs #674 and #690.

**Related Issue**

#336 - Background Tasks

**Proposed Changes**

This PR makes it possible for the FtResponse class to handle Background Tasks. Also documents the why, when, and how of Background Tasks.

**Types of changes**

**Checklist**

**Additional Information**


nbs/api/00_core.ipynb
```diff
@@ -3266,6 +3267,89 @@
...
+     "text": [
+      "Starting slow task\n",
+      "Finished slow task\n"
+     ]
+    }
+   ],
+   "source": [
+    "def my_slow_task():\n",
+    "    print('Starting slow task')    \n",
+    "    time.sleep(3)\n",
```
jph00, line 3296: This will make the whole notebook run and test slowly. How about using 0.001 instead?

nbs/api/00_core.ipynb
```diff
@@ -3266,6 +3267,89 @@
...
+      "Slept for 1s\n",
+      "Sleeping for 2s\n",
+      "Slept for 2s\n"
+     ]
+    }
+   ],
+   "source": [
+    "def increment(amount):\n",
+    "    print(f'Sleeping for {amount}s')    \n",
+    "    time.sleep(amount)\n",
```
jph00, line 3339: Maybe divide by 1000 here?

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "- Batch processes that can take a significant amount of time (bulk email or API calls)\n",
+    "- Any other process where the user can be notified later by email, websocket, webhook, or pop-up"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "::: {.callout-note}\n",
+    "Background tasks in FastHTML are built on Starlette's background tasks, but with some added sugar. Starlette's background task design is an easy-to-use wrapper around Python's async and threading libraries. Background tasks often make apps feel faster to the end user.\n",
```
jph00, line 28: They don't just *feel* faster -- they *are* faster! The user experiences the app returning immediately, rather than waiting for the bg task to complete.

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+   "metadata": {},
+   "source": [
+    "``` {.python filename=\"main.py\" code-line-numbers=\"true\"}\n",
+    "from fasthtml.common import *\n",
+    "from starlette.background import BackgroundTask\n",
+    "from time import time, sleep\n",
+    "\n",
+    "app, rt = fast_app()\n",
+    "\n",
+    "def counter(loops:int): # <1>\n",
```
jph00, line 57: No need for triple quotes for single-line docstrings

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "In this example we are attaching a task to FtResponse by assigning it via the background argument. When the page is visited, it will display 'Simple Background Task Example' almost instantly, while in the terminal it will slowly count upward from 0."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "``` {.python filename=\"main.py\" code-line-numbers=\"true\"}\n",
+    "from fasthtml.common import *\n",
+    "from starlette.background import BackgroundTask\n",
```
jph00, line 52: AFAICT you only use `sleep`, not `time`

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "```python\n",
+    "# main.py\n",
+    "from fasthtml.common import *\n",
+    "from fastlite import *\n",
```
jph00, line 148: This is already in fasthtml.common so no need to import it.

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "from starlette.background import BackgroundTask\n",
+    "import time\n",
+    "import httpx\n",
+    "\n",
+    "app, rt = fast_app()\n",
+    "\n",
+    "db = database(':memory:')\n",
+    "\n",
```
jph00, line 156: Place class on 1 line. No need for 3 separate callouts here IMO - it's clear enough and isn't specific to what you're teaching.

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "    task = BackgroundTask(task_submit, request_time=request_time) # <7>\n",
+    "    return FtResponse(resp_content, background=task) # <8>\n",
+    "\n",
+    "@rt\n",
+    "def show_tstamps(): return Ul(*[Li(t) for t in tstamps()]) # <9> \n",
+    "\n",
+    "@rt\n",
+    "def index():\n",
+    "    return Titled('Background Task Dashboard',\n",
+    "        P(Button('Press to call slow service', # <10> \n",
```
jph00, line 182: Use `hx_post=submit` here.

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "            hx_post='/submit', hx_target='#res')),\n",
+    "        H2('Responses from Tasks'),\n",
+    "        P('', id='res'),\n",
```
jph00, line 185: I think `map(Li, tstamps())` is nicer IMO - also I don't think you need `*` here IIRC.

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "Multiple background tasks on a background task are executed in order. In the case a task raises an exception, following tasks will not get the opportunity to be executed.\n",
+    ":::"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "```python\n",
+    "from fasthtml.common import *\n",
```
jph00, line 245: remove

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "from starlette.background import BackgroundTasks\n",
+    "\n",
+    "@rt\n",
+    "async def signup(email, username):\n",
```
jph00, line 249: I don't think you need all this - IIRC you can just return multiple bg tasks in your returned tuple.

nbs/explains/background_tasks.ipynb
```diff
@@ -0,0 +1,296 @@
...
+    "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Background tasks are not distributed task queues\n",
+    "\n",
+    "While background tasks often provide the user with a faster experience, the server itself isn't accelerated. This means processes are still happening, but are hidden from the user. So if a server is struggling under the load of a lot of user activity, so long as it isn't an issue with open HTTP connections consuming resources, background tasks won't help with server load.\n",
```
jph00, line 270: That's not necessarily true AFAICT. E.g if the server is struggling because a non-async process is blocking, then putting it in a bg task will help. So maybe worth mentioning the benefit that bg tasks are async would help here? And perhaps that spinning out a *process* manually can be a useful step between bg tasks and task queues, since it allows us to use multiple cores and avoid the GIL whilst still doing everything in fasthtml?

jph00 (changes requested):
This looks great! I've got a few minor requests to look at.

nbs/api/00_core.ipynb
pydanny: I used to have it longer but made it shorter. Still shorter we will go!

PR #701, opened 2025-04-03
change name in pyproject.toml

johnowhitaker:
Attempting to install with UV gives an error:

image

Minimal repro: https://colab.research.google.com/drive/1WWqJTaZAAq_CsCzmH602XKBpVYbDh2Pm?usp=sharing


johnowhitaker:
Wait scratch that this is generated by nbdev so we need to fix it there

PR #738, opened 2025-05-24
[PR] Allow any Mapping children to specify attributes. fix #737

### LINKED ISSUE #737: [FEATURE] Allow any Mapping for attribute dictionaries
If you'd like to discuss your feature idea first with the community (highly recommended!) please visit our [Discord channel](https://discord.gg/qcXvcxMhdP).

**Is your feature request related to a problem? Please describe.**
When passing in a custom Mapping it doesn't get detected as attributes. e.g. `Div(MyMapping(), "content")`

**Describe the solution you'd like**
Currently the code is using an `isinstance(x, dict)` check, I'd like to change it to `isinstance(x, Mapping)` to allow any mapping types.
https://github.com/AnswerDotAI/fasthtml/blob/ad42bb8fb70b47516da914aa5ea633e800c04191/fasthtml/components.py#L87

**Example code**

```python
# Your example code here
Div(MyMapping(), {'regular': 'dict'}, Span("content"))
```

**Similar implementations**
If available, provide links to similar features in other libraries:
1. [htpy](https://htpy.dev/usage/#attributes-as-dict) allows any Mapping, not just dicts.

**Problem solved**
Explain what problem this feature would solve and how it would benefit users of the library:
I'm creating a helper for datastar which allows constructing datastar attributes. Ideally this helper is easily usable with multiple html building frameworks. Rather than subclassing a dict, it subclasses Mapping to make things simpler. Currently it has to be used with Fasthtml via unpacking `Div(**data.on("click", "console.log('clicked');"))` but ideally it would work the same as a dictionary as well. `Div(data.on("click", "console.log('clicked');")` https://github.com/starfederation/datastar/pull/902

**Additional context**

**Confirmation**
Please confirm the following:
- [x] I have checked the existing issues and pull requests to ensure this feature hasn't been requested before.
- [x] I have read the project's documentation to ensure this feature doesn't already exist.

gazpachoking:
**Related Issue**
#737 

**Proposed Changes**

Dictionaries are allowed as a child of fasttags, which turns them into attributes. This change makes it so that any Mapping can be used for this, rather than just `dict`s. The purpose is to allow helpers to generate attributes that don't have to subclass the builtin `dict`.

**Types of changes**

**Checklist**

**Additional Information**

### DIFF
--- a/fasthtml/components.py
+++ b/fasthtml/components.py
@@ -19,7 +19,8 @@ __all__ = ['named', 'html_attrs', 'hx_attrs', 'hx_evts', 'js_evts', 'hx_attrs_an
 # %% ../nbs/api/01_components.ipynb
 from dataclasses import dataclass, asdict, is_dataclass, make_dataclass, replace, astuple, MISSING
 from bs4 import BeautifulSoup, Comment
-from typing import Literal, Optional
+from collections import UserDict
+from typing import Literal, Mapping, Optional
 
 from fastcore.utils import *
 from fastcore.xml import *
@@ -84,7 +85,7 @@ fh_cfg['auto_name']=True
 
 # %% ../nbs/api/01_components.ipynb
 def ft_html(tag: str, *c, id=None, cls=None, title=None, style=None, attrmap=None, valmap=None, ft_cls=None, **kwargs):
-    ds,c = partition(c, risinstance(dict))
+    ds,c = partition(c, risinstance(Mapping))
     for d in ds: kwargs = {**kwargs, **d}
     if ft_cls is None: ft_cls = fh_cfg.ft_cls
     if attrmap is None: attrmap=fh_cfg.attrmap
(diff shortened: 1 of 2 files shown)


fasthtml/components.py
```diff
@@ -19,7 +19,8 @@
 # %% ../nbs/api/01_components.ipynb
 from dataclasses import dataclass, asdict, is_dataclass, make_dataclass, replace, astuple, MISSING
 from bs4 import BeautifulSoup, Comment
-from typing import Literal, Optional
+from collections import UserDict
```
jph00, line 22: Looks like this is only used for a test/example, so it shouldn't be exported. To fix this, move this import into the cell that does not have a `#| export` header.

gazpachoking: Aha. I thought that was weird that the test import went into the actual source. Didn't see how that was working. Fixed.

PR #755, opened 2025-07-08
support x-forwarded-host as source for redirect

erikgaas:
In a situation where we have a reverse proxy redirect like example.com upon the actual hostname of localhost:8000 OAuth causes some issues where it still will want to redirect to port 8000.

I don't think this is the expected behavior we are looking for so if the x-forwarded-host exists I let oauth consume that info instead for the redirect.


jph00:
Great!

PR #761, opened 2025-07-23
Use data instead of json

erikgaas:
Fix OAuth2 token exchange to use form encoding instead of JSON

Changed httpx.post() to use `data=` instead of `json=` parameter to ensure 
requests are sent with `application/x-www-form-urlencoded` content type, 
as required by [RFC 6749 Section 4.1.3](https://tools.ietf.org/html/rfc6749#section-4.1.3) for OAuth2 token exchange endpoints.

This fixes authentication issues with Google OAuth2 and ensures compliance 
with other OAuth2 providers that strictly enforce the specification.


erikgaas:
@jph00 I needed this change in order to get oauth working for what I was working on. I didn't understand why and used AI to help me understand the problem. This seems like the right change, but I can't explain why so many other google client oauth applications work without this change.

PR #778, opened 2025-09-11
Correct afterware resp objects

erikgaas:
Followup to https://github.com/AnswerDotAI/fasthtml/pull/777

I dove into the internals for how these functions work. The change works, but I dont think it always returns the correct response object which may influence the behavior of the toast. 

For example:

```python
def toast_after(resp, req, sess):
    if sk in sess and (not resp or isinstance(resp, (tuple,FT,FtResponse))):
        sess['toast_duration'] = req.app.state.toast_duration
        req.injects.append(render_toasts(sess))
```

The current code always passes in None which appears to be what exists in the req object. updating the wreq object to use the resp object appears to correctly pass it in.

That way if toast receives a Redirect response, for example, then it knows not to do the toast according to the code above.

### DIFF
--- a/nbs/api/00_core.ipynb
+++ b/nbs/api/00_core.ipynb
@@ -3904,21 +3904,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "python3",
    "language": "python",
    "name": "python3"
-  },
-  "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.12.9"
   }
  },
  "nbformat": 4,


erikgaas:
Inputs to toast_after when not setting the wreq object:
```
after toast None <starlette.requests.Request object at 0x124e3a300> {'toast_duration': 5000, 'toasts': [('Successful Page Load!', 'success', False)]}
after toast None <starlette.requests.Request object at 0x124e39dc0> {'toast_duration': 5000}
```

Inputs to toast_after setting resp in the wreq object:
```
after toast <p>I love toast!</p> <starlette.requests.Request object at 0x165539eb0> {'toast_duration': 5000, 'toasts': [('Successful Page Load!', 'success', False)]}
after toast <starlette.responses.FileResponse object at 0x165477aa0> <starlette.requests.Request object at 0x1655399d0> {'toast_duration': 5000}
```

erikgaas:
I'll include this as a test in the notebook : )

jph00:
Thanks @erikgaas ! Can you think of any away to avoid having to create that test-specific pair of toast functions? Not a big deal, but would be nice if it were cleaner.

jph00:
BTW @erikgaas add a 'bug' or 'enhancement' label to PRs to ensure they get into the release notes. (I added this one).

Issue #668, opened 2025-03-05
[BUG] html2ft handling of white space


jsandeo:
**Describe the bug**
html2ft strips leading and trailing white space from text nodes, which might result in incorrect ft expressions.

**Minimal Reproducible Example**
Enter this HTML at https://h2f.answer.ai/ :

```html
Howdy <a href="https://answer.ai">answer</a> how are you?
```

**Expected behavior**
You should get
```python
'Howdy '
A('answer', href='https://answer.ai')
' how are you?'
```
but instead you get
```python
'Howdy'
A('answer', href='https://answer.ai')
'how are you?'
```

**Environment Information**
- fastlite version: 0.1.1
- fastcore version: 1.7.28
- fasthtml version: 0.12.1

**Confirmation**
- [✅] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [✅] I have provided a minimal reproducible example
- [✅] I have included the versions of fastlite, fastcore, and fasthtml
- [✅] I understand that this is a volunteer open source project with no commercial support.

**Screenshots**
![Image](https://github.com/user-attachments/assets/47064f54-5308-4a50-932c-b17cb4586c95)

johnowhitaker:
(Fixed in https://github.com/AnswerDotAI/fasthtml/pull/797 - thanks for spotting this)

PR #797, opened 2025-11-13
only strip newlines when parsing strings

johnowhitaker:
Related issue: https://github.com/AnswerDotAI/fasthtml/issues/668

Issue: inline spaces weren't preserved:

e.g. `"Howdy <a href="https://answer.ai">answer</a> how are you?"` -> "Howdy" (no trailing space). 

We changed it to only strip newlines from strings in _parse, and added a test. The original test failed, the HTML was functionally identical but different by one newline - independant of the trailing space bug. We now compare two soups of the HTML before vs after the round trip

### DIFF
--- a/nbs/api/01_components.ipynb
+++ b/nbs/api/01_components.ipynb
@@ -23,9 +23,7 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "8e2d405b",
-   "metadata": {
-    "time_run": "6:16:20p"
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "#| export\n",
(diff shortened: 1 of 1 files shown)


johnowhitaker:
@jph00 can you review when conveneint (I can explain when we next chat :) )

johnowhitaker:
To expand, this was a bug we picked up during an impromptu office hours. 
We first looked at the example from the issue and explored it in a dialog [here](https://share.solve.it.com/d/02b1472e0a9f8d3b6b8155e2a189a379). 

Then, once we were convinced it was a good idea, we moved to the fasthtml.components nb and made the tweak. We added an inline, informal test, but then noticed the 'tests' section right below and added a test for this case instead. It failed though, an errant newline unrelated to our change (I think). @erikgaas had the bright idea to check if bs4 could check equivalence of two soups, and we changed the test to use that instead of comparing the xml strings directly.

@jph00 I am not 100% sure this is right - can you think if there are cases where we definitely do want to `strip()` strings that are parts of elements? Or other issues with our fix? Do you think it makes sense to still strip newlines, or should weget rid of the `strip()`s entirely?

jph00:
lgtm! :) tbh I don't recall why the strip is there at all...

PR #804, opened 2025-12-03
Line defaults suppressed inheriting of stroke color from parent elements

### LINKED ISSUE #757: [BUG]  SVG Line should not have some default parameters
in FastHTML 0.12.21 in file ```svg.py``` the ```Line``` is defined with default values for stroke and stroke_width:
```python
def Line(x1, y1, x2=0, y2=0, stroke='black', w=None, stroke_width=1, **kwargs): ...
```
This results that using a line with only coordinates, will also auto generate:  ```stroke="black" stroke-width="1"```
This is wrong as it overrides attributes defined in a parent tag like ```<g>```. 
Eg. this will fail to achieve the result of having red and thick lines:
```python
G(stroke="red", stroke_width="20" )(
        Line(x1=0, y1=0, x2=100, y2=100),
        Line(x1=0, y1=0, x2=10, y2=100),
    )
```
Instead of desired outputs like:
```svg
<line x1="0" y1="0" x2="100" y2="100">
```
will generate lines that have unwanted attributes, that override the parent:
```svg
<line x1="0" y1="0" x2="100" y2="100" stroke="black" stroke-width="1">
```

erikgaas:
Fixes #757 

Line defaults suppressed inheriting of stroke color from parent elements.

### DIFF
--- a/nbs/api/05_svg.ipynb
+++ b/nbs/api/05_svg.ipynb
@@ -212,7 +212,20 @@
    "execution_count": null,
    "id": "7c4db30a",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<svg xmlns=\"http://www.w3.org/2000/svg\" viewbox=\"0 0 50 50\" height=\"50\" width=\"50\"><rect width=\"30\" height=\"30\" x=\"0\" y=\"0\" fill=\"blue\" rx=\"8\" ry=\"8\"></rect></svg>"
+      ],
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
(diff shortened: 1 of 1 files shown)


jph00:
Can you check this still seems to work ok?:
https://github.com/AnswerDotAI/fasthtml-example/blob/main/htmx/svgswap.py

And also that the docs render ok?:
https://www.fastht.ml/docs/api/svg.html

erikgaas:
> Can you check this still seems to work ok?: https://github.com/AnswerDotAI/fasthtml-example/blob/main/htmx/svgswap.py
> 
> And also that the docs render ok?: https://www.fastht.ml/docs/api/svg.html

Both seem to work fine.

jph00:
Thanks!

PR #805, opened 2025-12-03
Update docs on path resolution

erikgaas:
Path resolution is not well documented. I have added some added information to specify the special verbs used.


algal:
LGTM, and I learned something from it!

jph00:
Super! :D

Issue #775, opened 2025-08-27
[BUG] Example not work


analista01:
In the example (https://www.fastht.ml/docs/tutorials/by_example.html):

class Person:
    def __init__(self, name, age):
        self.name = name
        self.age = age

    def __ft__(self):
        return ['div', [f'{self.name} is {self.age} years old.'], {}]

p = Person('Jonathan', 28)
print(to_xml(Div(p, "more text", cls="container")))

Output not render the tag "div", show:

<div class="container">
['div', ['Jonathan is 28 years old.'], {}]more text</div>

analista01:
This example in the tutorial not work:

from fastcore.all import patch

@patch
def __ft__(self:Person):
    return Div("Person info:", Ul(Li("Name:",self.name), Li("Age:", self.age)))

show(p)

show message: AttributeError: 'NoneType' object has no attribute 'display'

78wesley:
I can confirm the first comment is an issue in version 0.12.28.

The `show()` function only works with Jupyter Notebooks.

```python
from fasthtml.common import *


class Person:
    def __init__(self, name, age):
        self.name = name
        self.age = age

    def __ft__(self):
        return ["div", [f"{self.name} is {self.age} years old."], {}]


p = Person("Jonathan", 28)
print(to_xml(Div(p, "more text", cls="container")))

```

Returns:
```
<div class="container">
['div', ['Jonathan is 28 years old.'], {}]more text</div>
```

erikgaas:
Agreed this is not correct syntax this should be. I've added a pr to update the docs.

```python
class Person:
    def __init__(self, name, age):
        self.name = name
        self.age = age

    def __ft__(self):
        return Div(f'{self.name} is {self.age} years old.')
```

Will close this issue for now unless I'm wrong and somehow this should be a supported syntax

PR #806, opened 2025-12-03
Fix ft documentation in "from example" guide.

erikgaas:
The current tutorial for ft tags says output can be 

```python
class Person:
    def __init__(self, name, age):
        self.name = name
        self.age = age

    def __ft__(self):
        return ['div', [f'{self.name} is {self.age} years old.'], {}]

p = Person('Jonathan', 28)
print(to_xml(Div(p, "more text", cls="container")))
```

Notice the list being returned by ft. I think this is not valid, so this is fixed in the docs.

Addresses confusion in #775

### DIFF
--- a/nbs/tutorials/by_example.ipynb
+++ b/nbs/tutorials/by_example.ipynb
@@ -2082,10 +2082,7 @@
    ]
   }
  ],
- "metadata": {
-  "solveit_dialog_mode": "learning",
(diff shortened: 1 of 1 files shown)


jph00:
Oh gosh yes that was a very old approach! I was rather proud however that at the least we were literally "just a `list`"! :D

PR #807, opened 2025-12-05
Bump toast z-index to 1090 so toasts appear above modals

erikgaas:
Toasts were using `z-index: 1000`, but UIkit modals (used by MonsterUI/FrankenUI) use `z-index: 1010`. This caused toasts to be hidden behind modals.

Bumped to `z-index: 1090` to ensure toasts always appear above modals while leaving room in the z-index scale for other elements.

Related: AnswerDotAI/MonsterUI#148, AnswerDotAI/solveit#1071

### DIFF
--- a/fasthtml/toaster.py
+++ b/fasthtml/toaster.py
@@ -7,7 +7,7 @@ tcid='fh-toast-container'
 sk = "toasts"
 toast_css = """
 #fh-toast-container {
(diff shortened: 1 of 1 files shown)


curtis-allan (approved):
Not sure how I haven't seen this conflict before, maybe they were really well hidden or I've just managed to avoid it all this time :/ Agree we'd want FastHTML toasts to take priority though so LGTM

jph00:
Thanks! :)

PR #810, opened 2025-12-12
Fix CommonMark 404 on index page

### LINKED ISSUE #795: Clicking CommonMark takes you to 404 url
On [the docs page](https://www.fastht.ml/docs/index.html) there's a `CommonMark` link, which I believe should yield a markdown version of the docs, but it links to the following NotFound url: `https://www.fastht.ml/docs/.md`

erikgaas:
The CommonMark link on the docs index page goes to .md instead of index.html.md, causing a 404.

The original fix script had the right logic — it uses data-original-href which Quarto preserves with the correct URL. However, it ran on DOMContentLoaded, which is before Quarto's JS mangles the href. So by the time the fix ran, the href was still correct, and then Quarto broke it afterward.

This PR adds a setTimeout so the fix runs after Quarto's JS completes.

Testing: Verified manually on the live site by running the fix logic in the browser console — confirmed the selector finds the element and the fix restores the correct href. I haven't figured out how to make local testing repro because it works fine in local Quarto builds.

Fixes #795

### DIFF
--- a/nbs/fix-commonmark.html
+++ b/nbs/fix-commonmark.html
@@ -1,8 +1,11 @@
 <script type="application/javascript">
 // Fix CommonMark link bug where Quarto JS changes index.html.md to .md
+// Use setTimeout to run after Quarto's JS has finished mangling the href
 document.addEventListener('DOMContentLoaded', function() {
-  document.querySelectorAll('a[href=".md"]').forEach(link => {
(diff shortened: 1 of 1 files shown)


jph00:
Thanks @erikgaas. FYI this is a problem in all Quarto/nbdev projects - so even better would be to fix it at the source, if possible!

PR #812, opened 2025-12-12
Add note about Post/Redirect/Get pattern in by_example tutorial

### LINKED ISSUE #389: [BUG] FastHTML by Example should use a redirect for POST instead of render
**Describe the bug**
This is more a "bug"/question about the tutorial. I was following along and noticed when I got to the [WebPage->Web App](https://docs.fastht.ml/tutorials/by_example.html#web-page---web-app) section that things did not work in the browser. 

1. There seemed to be some loop (that I can not repo) that caused the page to keep adding the last item I submitted
2. After you add an items and get back to the home page, if you refresh the page it wants to do another form submission 

Feels like both issues are because the POST renders the home page instead of redirecting

```
@app.post("/")
def add_message(data:str):
    messages.append(data)
    return home()
```

vs

```
@app.post("/")
def add_message(data: str):
    messages.append(data)
    return RedirectResponse("/", status_code=302)
```

**Expected behavior**
If I am follow a demo, the demo produces best practice web patterns. I had to go looking through GitHub example repos to even find that RedirectResponse is the right way to do this. 

**Environment Information**
- fastlite version: "0.0.9"
- fastcore version:"1.7.3"
- fasthtml version:"0.5.1"

**Confirmation**
Please confirm the following:

**Additional context**
Note: problem #1 might have been because I was using an earlier version of fasthtml and recently updated, but #2 to exists still.

erikgaas:
Adds a brief note before the HTMX section explaining that refreshing after a POST can cause "Resubmit form data?" prompts, and that `Redirect('/')` is the traditional fix (the [Post/Redirect/Get](https://en.wikipedia.org/wiki/Post/Redirect/Get) pattern).

Also notes this isn't needed with HTMX since submissions happen via AJAX — which is a nice segue into why HTMX is useful.

Fixes #389

### DIFF
--- a/nbs/tutorials/by_example.ipynb
+++ b/nbs/tutorials/by_example.ipynb
@@ -682,6 +682,14 @@
   },
   {
    "cell_type": "markdown",
+   "id": "prg-note",
+   "metadata": {},
+   "source": [
+    "> **Note:** With traditional form submissions, refreshing the page after a POST will prompt \\\"Resubmit form data?\\\" — potentially causing duplicate entries. The standard fix is `return Redirect('/')` instead of `return home()`, known as the [Post/Redirect/Get](https://en.wikipedia.org/wiki/Post/Redirect/Get) pattern. With HTMX (covered next), this isn't needed since submissions happen via AJAX."
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "0e7066db",
    "metadata": {},
    "source": [


jph00:
Thanks @erikgaas . Minor nit - but I'm not sure "ajax" is the best term here. That is specific to `XMLHttpRequest` requests, but nowadays most APIs, including the upcoming HTMX v4, use `fetch`.

PR #813, opened 2025-12-14
Add Apple sign in

erikgaas:
Adding sign in with apple. It is slightly different because it requires parsing a p8 file and a post request for the redirect if you want name and email.

Adding PyJWT as a req. It's small and fairly ubiquitous, but I'm down to reconsider.

Fun fact apple sign in does not work on localhost which is super annoying, but you can use solveit easily to test it out : )


jph00:
Nice one @erikgaas ! :) If you move jwt to being a dev requirement, and then only import it *inside* the functions where it's used, then it doesn't add a dep for everyone.

erikgaas:
> Nice one @erikgaas ! :) If you move jwt to being a dev requirement, and then only import it _inside_ the functions where it's used, then it doesn't add a dep for everyone.

Nice! good idea. changed

PR #823, opened 2026-01-16
Fix return type annotation bug

### LINKED ISSUE #814: [BUG] Tighter type annotations on route return types triggers starlette error
**Description**
I'm moving from not have any type annotations to actually having some. However, adding stronger type annotation on route method return types triggers a starlette error.

```python
app, rt = fast_app()

@rt("/works")
def any() -> Any:
    return Titled("A Title", H2("An H2"))

@rt("/errors")
def titled() -> Titled:
    return Titled("A Title", H2("An H2"))

serve()
```
/works works. /errors gives me the following starlette error:
```
....*much deep stack*...

  File "/Users/peter/Repository/10-19 Development/10 Development/POCs/python-code-quality/.venv/lib/python3.11/site-packages/starlette/routing.py", line 76, in app
    await response(scope, receive, send)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'tuple' object is not callable
```

**Expected behavior**
I was hoping that adding a more focused type annotation would not trigger a runtime error; is starlette doing something more fancy underneath?

I've tried various alternatives (e.g. "Titled" and type-vars) and still had the same problem. Also happens irrespective of the actual component return, i.e. a method that returns a Div that's typed as such also doesn't work.

Is the section in the FAQ wrt PEP-8 and ruff indicative that the fastapi-coding style not only eschews type annotations but that they actively don't work?

**Environment Information**
- fastlite version: 0.2.2
- fastcore version: 1.9.2
- fasthtml version: 0.12.36

**Confirmation**
Please confirm the following:
- [X] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [X] I have provided a minimal reproducible example
- [X] I have included the versions of fastlite, fastcore, and fasthtml
- [X] I understand that this is a volunteer open source project with no commercial support.

erikgaas:
Fixes #814.

Previously, return type annotations like `-> Titled` would cause a `TypeError: 'tuple' object is not callable` because FastHTML tried to call the annotation as a constructor.

The fix: only use the return type annotation for wrapping when it's actually a `Response` subclass. This makes sense because the wrapping code passes `status_code` as a keyword argument, which only Response subclasses accept.

### DIFF
--- a/nbs/api/00_core.ipynb
+++ b/nbs/api/00_core.ipynb
@@ -3963,10 +3963,7 @@
    "source": []
   }
  ],
- "metadata": {
-  "solveit_dialog_mode": "learning",
-  "solveit_ver": 2
- },
+ "metadata": {},
  "nbformat": 4,
  "nbformat_minor": 5
 }


fasthtml/core.py
```diff
@@ -445,7 +445,7 @@ def _resp(req, resp, cls=empty, status_code=200):
     "Create appropriate HTTP response from request and response data"
     if not resp: resp=''
     if hasattr(resp, '__response__'): resp = resp.__response__(req)
-    if cls in (Any,FT): cls=empty
+    if not (isinstance(cls, type) and issubclass(cls, Response)): cls=empty
```
erikgaas: The Any, and FT checks are covered in isisntance type check. the type checking is required or else issubclass Response will throw an error.

erikgaas:
I could also see this however as suppressing errors if someone expects some type of behavior. so we could throw an error? but this seems along the lines of what was intended for now

jph00:
Good idea! :)

PR #824, opened 2026-01-29
add optional client arg to HTMX for persistent tests

johnowhitaker:
When testing FastHTML apps with authentication in notebooks, the default HTMX function creates a fresh TestClient for each render. This means session cookies (like auth) don't persist, making it difficult to preview authenticated content. This change allows one to pass in an existing TestClient to use, allowing for preview of things that require session state.

Worked example: https://share.solve.it.com/d/d19bed2aa88ed7476a03b821a29caef7

### DIFF
--- a/fasthtml/jupyter.py
+++ b/fasthtml/jupyter.py
@@ -118,7 +118,7 @@ from starlette.testclient import TestClient
 from html import escape
 
 # %% ../nbs/api/06_jupyter.ipynb #89b8984b
-def HTMX(path="/", host='localhost', app=None, port=8000, height="auto", link=False, iframe=True):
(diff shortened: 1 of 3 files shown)


fasthtml/jupyter.py
```diff
@@ -137,7 +137,8 @@ def HTMX(path="/", host='localhost', app=None, port=8000, height="auto", link=Fa
         route = f'/{unqid()}'
         res = path
         app.get(route)(lambda: res)
-        page = TestClient(app).get(route).text
+        cli = client if client is not None else TestClient(app)
```
johnowhitaker: This is the change: can now use an existing client if provided

johnowhitaker:
@jph00 :)

jph00:
Thank you! :)

PR #826, opened 2026-02-04
Fix CSS syntax in by_example.ipynb

egrim:
An example of applying custom CSS in `by_example.ipynb` that adds a custom style tag to `hdrs` contains an errant comma where a semicolon should be used.  This prevents the custom font-family that's specified from actually being used when served to a browser.


jph00:
Well spotted!

PR #829, opened 2026-02-07
fix: allow serve() to detect any entry point function name

### LINKED ISSUE #816: [BUG] `serve()` auto-detection fails silently if entry point function is not named `main`
### **Describe the bug**

The `serve()` function attempts to auto-detect the `appname` (module path) to enable `uvicorn` reloading. However, the detection logic in `api/00_core.ipynb` (exported to `fasthtml/common.py`) contains a hardcoded check that requires the calling function to be named exactly `"main"`.

If a user defines a console script entry point in `pyproject.toml` pointing to a function with a different name (e.g., `start()` or `run_app()`), the `appname` detection fails, returning `None`. Consequently, `uvicorn.run` is never called, and the program exits silently without error.

### **Minimal Reproducible Example**

**1. Python File (`myapp.py`)**

```python
from fasthtml.common import serve, FastHTML

app = FastHTML()

# This function name causes the failure
def start_app():
    serve()

# If you rename this to 'def main():', it works.

```

**2. pyproject.toml**

```toml
[project.scripts]
# Entry point pointing to a non-main function
myserver = "myapp:start_app"

```

**3. Execution**
Run the console command:

```bash
$ myserver
# Result: Process exits immediately. Server does not start.

```

### **Expected behavior**

`serve()` should correctly detect the module name via `inspect` regardless of the specific name of the entry point function. The logic should rely on the stack frame context (`bk.f_back.f_globals.get('__name__') == '__main__'`) rather than `code.co_name == 'main'`.

### **Environment Information**

* fastlite version: 0.2.3
* fastcore version: 1.9.4
* fasthtml version: 0.12.36

### **Confirmation**

* [x] I have read the FAQ ([https://docs.fastht.ml/explains/faq.html](https://docs.fastht.ml/explains/faq.html))
* [x] I have provided a minimal reproducible example
* [x] I have included the versions of fastlite, fastcore, and fasthtml
* [x] I understand that this is a volunteer open source project with no commercial support.

### **Additional context**

The issue is located in the `serve` function logic.

**Current Logic (causes failure):**

```python
elif code.co_name=='main' and bk.f_back.f_globals.get('__name__')=='__main__': 
    appname = inspect.getmodule(bk).__name__

```

**Suggested Fix:**
Remove the `code.co_name=='main'` check. If the caller's globals indicate it was invoked by the `__main__` entry point wrapper, the function name is irrelevant.

```python
elif bk.f_back.f_globals.get('__name__')=='__main__': 
    appname = inspect.getmodule(bk).__name__

```

thecaptain789:
Fixes #816.

The `serve()` function previously required the calling function to be named exactly `'main'`. This restriction is unnecessary since the key check is whether we're running from `__main__`, not the function name.

**Before:** Entry points like `myapp:start_app` would fail silently.

**After:** Any function name works as long as it's called from `__main__`.

Changes:
- Removed the `code.co_name=='main'` check from appname detection
- Removed unused `code` variable

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -714,10 +714,9 @@ def serve(
     "Run the app in an async server, with live reload set as the default."
     bk = inspect.currentframe().f_back
     glb = bk.f_globals
-    code = bk.f_code
     if not appname:
         if glb.get('__name__')=='__main__': appname = Path(glb.get('__file__', '')).stem
-        elif code.co_name=='main' and bk.f_back.f_globals.get('__name__')=='__main__': appname = inspect.getmodule(bk).__name__
+        elif bk.f_back.f_globals.get('__name__')=='__main__': appname = inspect.getmodule(bk).__name__
     import uvicorn
     if appname:
         if not port: port=int(os.getenv("PORT", default=5001))


jph00:
This is an nbdev project so changes need to be made in the ipynb files, then exported.

PR #837, opened 2026-03-02
Cache request body if content-type header is not `multipart/form-data`

curtis-allan:
This PR ensure `await request.body()` works within routes, regardless of params/ annotations.

It works by calling `await req.body()` once within `parse_form`. Starlette automatically caches the body bytes data once this has been called (within req._body), and automatically uses the cached value for subsequent `req.form()` or `req.json()` calls. This also works for further `req.body()` calls too, which will use the cached body data and avoid `Stream Consumed` issues.

The `multipart/form-data` check is to avoid possibly overloading memory with large uploaded files (which require `multipart/form-data` content type to be accessed server-side correctly), which means file uploads will be accessed via the stream directly instead of cached data.

Trade-off IMO is worth it here - to avoid possible server issues, but not have the body accessible via direct `await req.body()` calls in routes when the content-type is `multipart/form-data` (still accessible via the `body` param if needed!)

Also fixes a breaking `to_xml` change in fastcore (`indent` and further params need converting to `kwargs` now) and ensures `nbdev-test` works correctly

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -145,15 +145,14 @@ def form2dict(form: FormData) -> dict:
 async def parse_form(req: Request) -> FormData:
     "Starlette errors on empty multipart forms, so this checks for that situation"
     ctype = req.headers.get("Content-Type", "")
-    if not ctype.startswith('multipart'): await req.body()
-    if ctype=='application/json': return await req.json()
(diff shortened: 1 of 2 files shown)


jph00:
Thanks Curtis! :)

PR #839, opened 2026-03-02
fix: pass indent as keyword arg in _to_xml call

Jhutty:
---
name: Pull Request
title: '[PR] Fix: pass indent as keyword arg in _to_xml call'
labels: ''
assignees: ''

---

**Related Issue**
No existing issue. `fh_cfg.indent` being `True` causes `TypeError: sequence item 1: expected str instance, bool found` when rendering any route.

**Proposed Changes**
In `fasthtml/core.py` line 371, `indent` is passed as a positional argument to `to_xml()`. Since `to_xml` uses `*elms`, this causes `indent` to be treated as an HTML element to render. Changed to `indent=indent` (keyword argument).

**Types of changes**

**Checklist**
- [ ] I am aware that this is an nbdev project, and I have edited, cleaned, and synced the source notebooks instead of editing .py or .md files directly.

**Additional Information**
The bug occurs when `fh_cfg.indent` is set to `True` (boolean). The `to_xml(*elms, ...)` signature captures it as an element, then `'\n'.join(...)` fails because it encounters a `bool` instead of a string.

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -368,7 +368,7 @@ def _to_xml(req, resp, indent):
     "Convert response to XML string with target URL resolution"
     resp = _apply_ft(resp)
     _find_targets(req, resp)
-    return to_xml(resp, indent)
+    return to_xml(resp, indent=indent)
 
 # %% ../nbs/api/00_core.ipynb #f1e3ed2d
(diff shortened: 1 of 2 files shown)


jph00:
Thanks! Already fixed on our end :)

PR #841, opened 2026-03-05
Fix _fix_anno mangling bare `list` annotations

jackhogan:
Bare `list` annotations (without type params) causes silent data corruption — e.g. `["foo", "bar"]` becomes `["b", "a", "r"]` — and we get an `IndexError` for empty lists.

image

e.g. the above works with `list[str]` but doesn't with just `list`

### DIFF
--- a/nbs/api/00_core.ipynb
+++ b/nbs/api/00_core.ipynb
@@ -134,7 +134,7 @@
        "datetime.datetime(2026, 3, 9, 14, 0)"
       ]
      },
-     "execution_count": 10,
+     "execution_count": null,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -155,7 +155,7 @@
        "True"
       ]
      },
-     "execution_count": 11,
+     "execution_count": null,
      "metadata": {},
      "output_type": "execute_result"
     }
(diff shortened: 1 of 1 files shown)


jph00:
@jackhogan definitely agree the current behavior is not good - although I'm wondering if the *correct* behavior for non-specialised `list` would be the same as having no type at all? Because without the generic sub-type, we don't know what it's meant to be a list of. Or do we just assume it means "list of strs"?

If we want the latter, we should probably also support subclasses of both list and tuple? `is list` is perhaps a bit too specific.

jackhogan:
Good point @jph00! What do you think of this revised approach:

- **Bare `list`/`tuple` now warns and returns `None`** — consistent with how unannotated params are handled. Uses `issubclass(t, (list, tuple))` rather than `is list`, so subclasses are covered too.
- **Validate at declaration-time** — factored out the annotation check logic to be used at route declaration time (in `_endp`) so the user sees the warning straight away when they define the endpoint, and also at runtime (in `_find_p`). 

I also found a different bug while investigating: `Optional[list[str]]`, `list[str] | None`, and other Union-wrapped list types were *also* silently mangling data — `['foo', 'bar']` → `['b', 'a', 'r']`. This was because after unwrapping the Union in `_fix_anno`, `origin` was stale (still `Union` instead of `list`), so the `_mk_list` path was skipped. Fixed by re-deriving `origin` after the Union unwrap.

jph00:
Thanks @jackhogan :)

PR #854, opened 2026-03-27
fix: prevent query params leaking into typed session dict

### LINKED ISSUE #845: [BUG] Typehinting session parameter pulls in query parameters
**Describe the bug**
If you type hint the session parameter in a route function query parameters from the request get added to the session. Remove the type hinting and the query parameters don't get added.

In v0.12.44, the change "Add all HTTP request param handling to websockets too" refactored the parameter handling functions. The key problematic line introduced is in _find_ps:

```python
async def _find_ps(conn, data, hdrs, params):
    if conn.query_params: data |= dict(conn.query_params)  # <-- merges query params into data
    ...
```

This merges all query parameters into the data dict before any handler parameters are resolved. The problem occurs in _find_p when it processes a session parameter — if the session type is treated as a body type (dict or similar), the contaminated data dict (now containing query params) gets used to populate it.

The session then ends up containing all query parameters, which gets serialized into the cookie by Starlette's SessionMiddleware, blowing past the cookie size limit and causing the session to be cleared on every request.

Provide a minimal code snippet that reproduces the issue. This is crucial for us to understand and fix the bug quickly.

```python
from fasthtml.common import APIRouter, Request

router = APIRouter()

@router("/broken", ["GET"])
def get_video(request: Request, session: dict):
    # session will contain any query parameters from the request.

@router("/works", ["GET"])
def get_video(request: Request, session):
    # session will not contain query parameters from the request.
```

**Expected behavior**
I wouldn't expect typing the session parameter to change what's provided in the session parameter.

**Environment Information**
- fastlite version:0.2.4
- fastcore version:1.12.26
- fasthtml version:>=0.12.44

**Confirmation**
Please confirm the following:

temrjan:
## Summary

Fixes #845

When a session parameter is type-hinted as `dict`, query parameters from the request leak into the session, eventually causing the session cookie to exceed its size limit and get cleared on every request.

## Root Cause

In `_find_p`, the check `if anno is dict: return data` (line 201) runs **before** the session name check on line 203. Since `_find_ps` merges query params into `data` via `data |= dict(conn.query_params)`, returning `data` for `session: dict` includes all query parameters.

```python
# Before (broken order):
if anno is dict: return data          # ← catches session: dict, returns contaminated data
if _is_body(anno):
    if 'session'.startswith(arg.lower()): return conn.scope.get('session', {})  # ← never reached
```

## Fix

Move the session name check before `anno is dict`:

```python
# After (correct order):
if 'session'.startswith(arg.lower()) and _is_body(anno): return conn.scope.get('session', {})
if anno is dict: return data
if _is_body(anno):
    return await _from_body(conn, p, data)
```

## Test Plan

- [x] Fixed `test_get_toaster_with_typehint` (was failing on main)
- [x] Added `test_session_dict_no_query_param_leak` regression test
- [x] All 6 tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -198,9 +198,9 @@ async def _find_p(conn, data, hdrs, arg:str, p:Parameter):
         if issubclass(anno, Starlette): return conn.scope['app']
         if issubclass(anno, HTTPConnection): return conn
         if issubclass(anno, State): return conn.scope['app'].state
+        if 'session'.startswith(arg.lower()) and _is_body(anno): return conn.scope.get('session', {})
         if anno is dict: return data
         if _is_body(anno):
-            if 'session'.startswith(arg.lower()): return conn.scope.get('session', {})
             return await _from_body(conn, p, data)
     if (msg := _check_anno(arg, anno)): return warn(msg)
     # Special param names with no annotations
--- a/tests/test_toaster.py
+++ b/tests/test_toaster.py
@@ -46,10 +46,8 @@ def test_ft_response():
     assert 'Toast FtResponse' in res.text
 
 def test_get_toaster_with_typehint():
-    res = cli.get('/see-toast-with-typehint', follow_redirects=False)
-    assert 'Toast get' in res.text
-
-    res = cli.get('/see-toast-with-typehint', follow_redirects=True)
+    cli.get('/set-toast-get', follow_redirects=False)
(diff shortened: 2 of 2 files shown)


jph00:
Thanks. This is an nbdev project, so changes need to be made there, using coding style consistent with the rest of the project..

PR #853, opened 2026-03-27
Fix ws header lookup key mismatch

dienhoa:
---
name: Pull Request
about: Propose changes to the codebase
title: '[PR] '
labels: ''
assignees: ''

---

**Related Issue**
https://github.com/AnswerDotAI/fasthtml/issues/852
**Proposed Changes**
Fixes https://github.com/AnswerDotAI/fasthtml/issues/852

In `_wrap_ws`, replace the plain dict with Starlette's Headers and keep hyphenated keys instead of converting to underscores. This makes WS headers behave the same as HTTP headers, so `_find_p`'s `snake2hyphens` lookup works correctly.

**Types of changes**
What types of changes does your code introduce? Put an `x` in all the boxes that apply:
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

**Checklist**
Go over all the following points, and put an `x` in all the boxes that apply:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] I am aware that this is an nbdev project, and I have edited, cleaned, and synced the source notebooks instead of editing .py or .md files directly.

I also added a new test related to this in the nb:
```python
def on_receive(hx_trigger:str): return f"trigger: {hx_trigger}"
cli = TestClient(Starlette(routes=[WebSocketRoute('/', _ws_endp(on_receive))]))
with cli.websocket_connect('/') as ws:
    ws.send_text(json.dumps({"HEADERS": {"Hx-Trigger": "my-btn"}}))
    data = ws.receive_text()
    assert data == 'trigger: my-btn', data
```

@jph00

Issue #852, opened 2026-03-27
[BUG]  WS cannot resolve HTMX header parameters


dienhoa:
**Describe the bug**
After the unification of `_find_wsp` into `_find_p` (https://github.com/AnswerDotAI/fasthtml/commit/3583106552562884fa0617482aa91ee6337fa6d8), WS routes with (`@app.ws("/ws")`) can't resolve htmx header parameters (e.g `hx_trigger`). 

`_wrap_ws` stores headers as a plain dict with underscore keys (e.g `hx_trigger`), but `_find_p` looks them up via `snake2hyphens` → `Hx-Trigger`. Plain dicts require an exact key match, so the lookup fails

**Minimal Reproducible Example**
This example doesn't show `trigger: textid` after clicking the button https://github.com/AnswerDotAI/fasthtml-example/blob/main/htmx/ws.py

```python
from fasthtml.common import *
app, rt = fast_app(exts='ws')
@rt("/")
def get():
    return Div(
        P(ws_send=True, id="textid", name="textname", hx_ext="ws", ws_connect="/ws")(
            "Click Me"),
        P("Text to Update", id="dest")
    )
@app.ws("/ws")
async def ws(ws, send, hx_trigger:str):
    return Div(f"trigger: {hx_trigger}", id="dest", hx_swap_oob="true")
serve()
```

**Expected behavior**
`trigger: textid` after clicking on the button

**Environment Information**
Please provide the following version information:
- fastlite version: 0.2.4
- fastcore version: 1.12.31
- fasthtml version: 0.13.1

**Confirmation**
Please confirm the following:
- [x] I have read the FAQ (https://docs.fastht.ml/explains/faq.html)
- [x] I have provided a minimal reproducible example
- [x] I have included the versions of fastlite, fastcore, and fasthtml
- [x] I understand that this is a volunteer open source project with no commercial support.

PR #856, opened 2026-04-02
Fix WebSocket crash when HTMX sends null header values

### LINKED ISSUE #855: [BUG]
**Describe the bug**
Since v0.13.2 (PR #853), `_wrap_ws` in `core.py` wraps WebSocket `HEADERS` in Starlette's `Headers()` object:

```python
hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items()})
```

Starlette's `Headers()` constructor calls `.encode("latin-1")` on every value. When HTMX's WebSocket extension sends messages with `null` header values (e.g. `HX-Trigger: null`, which is normal when there is no triggering element), this crashes with:

```
AttributeError: 'NoneType' object has no attribute 'encode'
```

Prior to #853, headers were stored in a plain dict which handled `None` values without issue.

```python
from fasthtml.common import *

app, rt = fast_app(exts=["ws"])

@rt("/")
def get():
    return Div(
        Form(
            Input(name="msg", placeholder="Type here..."),
            id="form",
            ws_send=True,
        ),
        Div(id="output"),
        hx_ext="ws",
        ws_connect="/ws",
    )

@app.ws("/ws")
async def _(data, send):
    await send(Div(f"Got: {data.get('msg', '')}", id="output"))

serve()
```

Steps to reproduce:
1. Run the app and open it in a browser
2. Type text and press Enter to submit the form
3. Observe the server crash with `AttributeError: 'NoneType' object has no attribute 'encode'`

This occurs because HTMX includes headers like `HX-Trigger` with `null` values in the WebSocket JSON payload:

```json
{
  "msg": "hello",
  "HEADERS": {
    "HX-Request": "true",
    "HX-Trigger": null,
    "HX-Trigger-Name": null,
    "HX-Target": "output"
  }
}
```

**Expected behavior**
WebSocket messages with `null` header values should be handled gracefully without crashing. The fix is to filter out `None` values before constructing the `Headers` object:

```python
hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items() if v is not None})
```

**Environment Information**
- fastlite version: 0.2.4
- fastcore version: 1.12.34
- fasthtml version: 0.13.3

**Confirmation**

**Additional context**
The regression was introduced in PR #853 (commit d17be3c), which changed the header handling from a plain dict to Starlette `Headers()` to fix header key lookup (#852). The fix for #852 is correct, but `None` filtering was missed.

Current workaround: intercept `htmx:wsConfigSend` on the client side to replace `null` values with empty strings before sending.

aleph-ra:
**Related Issue**
Fixes #855

**Proposed Changes**
Filter out `None` values before constructing Starlette `Headers()` in `_wrap_ws` (`core.py:270`). This fixes the `AttributeError: 'NoneType' object has no attribute 'encode'` crash that occurs when HTMX's WebSocket extension sends `null` header values (e.g. `HX-Trigger: null`).

```python
# Before
hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items()})

# After
hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items() if v is not None})
```

This preserves the case-insensitive header lookup behavior introduced in #853 while gracefully handling `null` values.

**Types of changes**

**Checklist**

**Additional Information**
Regression introduced in #853 (commit d17be3c). The fix for #852 correctly switched from a plain dict to Starlette `Headers()` for case-insensitive lookup, but missed that HTMX sends `null` values for headers like `HX-Trigger` and `HX-Trigger-Name` when there is no triggering element. Starlette's `Headers()` constructor calls `.encode("latin-1")` on every value, which raises `AttributeError` on `None`.

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -267,7 +267,7 @@ async def _handle(f, *args, **kwargs):
 
 # %% ../nbs/api/00_core.ipynb #ad0f0e87
 async def _wrap_ws(ws, data, params):
-    hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items()})
+    hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items() if v is not None})
     return await _find_ps(ws, data, hdrs, params)
 
 # %% ../nbs/api/00_core.ipynb #dcc15129
--- a/nbs/api/00_core.ipynb
+++ b/nbs/api/00_core.ipynb
@@ -1083,7 +1083,7 @@
    "source": [
     "#| export\n",
     "async def _wrap_ws(ws, data, params):\n",
-    "    hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items()})\n",
+    "    hdrs = Headers({k.lower():v for k,v in data.pop('HEADERS', {}).items() if v is not None})\n",
     "    return await _find_ps(ws, data, hdrs, params)"
    ]
   },


jph00:
Thanks! :)

PR #863, opened 2026-04-17
add ratelimiting decorator

RensDimmendaal:
**This PR depends on:** #867

This PR adds a token-bucket rate limiter for FastHTML routes. I figured we needed something like this for the MagicKey PR to avoid abuse of the routes that send emails and I saw it was a common pattern for other web app frameworks. What's different about this implementation is that I've made it easy to re-use the request parameters as a `key` in the `limiter`.

I didn't add a separate explainer doc page, because the source doc page is quite extensive and builds up nicely. But i can add one if you like.

Example usage:

```py
 from fasthtml.common import *
 from fasthtml.ratelimit import limiter

 app, rt = fast_app()

 @rt(before=limiter('5/m'))
 def index(): return 'ok'

 @rt(before=limiter('100/h',key='user_id'))
 def api(user_id: str): return 'data'
```

**Types of changes**

**Checklist**

**Additional Information**

The changes to nbs 00 and 01 are due to the latest nbdev-version and not related to this PR.


fasthtml/ratelimit.py
```diff
@@ -0,0 +1,83 @@
...
+    "Get client IP from `X-Forwarded-For` header, falling back to `req.client.host`"
+    return req.headers.get('x-forwarded-for', '').split(',')[0].strip() or (req.client and req.client.host) or ''
+
+# %% ../nbs/api/07_ratelimit.ipynb #75b5b395
+class Limiter:
+    "Rate limiter for FastHTML routes"
+    def __init__(self, app):
+        self._req_var = ContextVar('limiter_req')
+        async def _store(req): self._req_var.set(req)
+        app.before.insert(0, _store)
```
RensDimmendaal: Ive been wondering if we could do this in a cleaner way. The issue is that `req` needs to be available in the decorator even when the route doesnt have req as a param.
If we can do that then Limiter can just be a function.

RensDimmendaal: One way could do it is in`core.py` create this global `current_req` ContextVar...but I'm not sure if thats agood or a really bad idea...Curious what you think!

```diff
+current_req = ContextVar('fh_current_req', default=None)
+
+@contextmanager
+def req_ctx(req):
+    token = current_req.set(req)
+    try: yield
+    finally: current_req.reset(token)

 @patch
 def _endp(self:FastHTML, f, body_wrap):
     "Create endpoint wrapper with before/after middleware processing"
     sig = signature_ex(f, True)
     for n,p in sig.parameters.items(): (msg:=_check_anno(n,p.annotation)) and warn(msg)
     async def _f(req):
+        with req_ctx(req):
            ...  # rest as before
```

RensDimmendaal: Resolved using the new `route(before=...)` cleans this up nicely

nbs/api/07_ratelimit.ipynb
RensDimmendaal, line 1: @jph00 I wasnt sure how to best handle the default key-ing on IP, because dependent on the reverse proxy, cloudflare and uvicorn setup.

I've made it such that it'll work for local dev and for our common deployment scenario of serving with uvicorn behind Caddy. 

I've added this note in the docs, and clarified when `client_ip` should (not) be used.

But I'm a bit unsure if this is the way to go...maybe it's too opinionated and we shouldnt make client_ip the default?

fasthtml/ratelimit.py
```diff
@@ -0,0 +1,74 @@
...
+        self.rate = max_reqs / window_secs
+        self.buckets = {}
+    def __repr__(self): return f'TokenBucket({self.max_reqs}, {self.window_secs})'
+
+    def _prune(self):
+        cutoff = time.time() - self.window_secs
+        self.buckets = {k:(t,ts) for k,(t,ts) in self.buckets.items() if ts > cutoff}
+    def wait(self, key):
+        "Return 0 if allowed, else seconds to wait"
+        self._prune()
```
comhar, line 43: Instead of running prune every time wait is called, perhaps we could check if the bucket has exceeded a certain number of keys or only prune on the nth request?

jph00:
Thanks @RensDimmendaal . Sorry for the delay - been thinking about this for a while. I'm not 100% sure we want this API, but if we do, binding it to IP probably isn't that useful. CGNAT is super common nowadays (e.g mobile, India, ...), and schools/companies/etc share an IP, etc - so you'd need big enough limits that the IP constraints wouldn't be very meaningful. Also, x-forwarded-for is easily spoofed. So I think if you're going to have an API like this, it should be for your whole service. I.e put in limits higher than you'd imagine you'd ever hit, so that if they ever are, it would be a disaster. (It's important to have it set up to notify you when that happens too.)

We can chat more f2f, but in general IMO this kinds of systems should have:
- Multiple customisable aggregation types (session cookie; user; email domain; IP; region; system-wide, etc)
- Multiple thresholds/actions (warn; temp block; block-until-admin review; full shutdown)
- Handle multiple backends sharing a single limit system
I can explain how things were done at FastMail if you like - we had a great system that which I really liked. For passkey we might want to find a way to simply call out to a limit aggregator instead of building something in.

fasthtml/ratelimit.py
```diff
@@ -0,0 +1,74 @@
...
+        now = time.time()
```
comhar, line 44: Should we use `time.monotonic`?

RensDimmendaal:
> Thanks @RensDimmendaal . Sorry for the delay - been thinking about this for a while. I'm not 100% sure we want this API, but if we do, binding it to IP probably isn't that useful. CGNAT is super common nowadays (e.g mobile, India, ...), and schools/companies/etc share an IP, etc - so you'd need big enough limits that the IP constraints wouldn't be very meaningful. Also, x-forwarded-for is easily spoofed. So I think if you're going to have an API like this, it should be for your whole service. I.e put in limits higher than you'd imagine you'd ever hit, so that if they ever are, it would be a disaster. (It's important to have it set up to notify you when that happens too.)
> 
> We can chat more f2f, but in general IMO this kinds of systems should have:
> 
> * Multiple customisable aggregation types (session cookie; user; email domain; IP; region; system-wide, etc)
> * Multiple thresholds/actions (warn; temp block; block-until-admin review; full shutdown)
> * Handle multiple backends sharing a single limit system
>   I can explain how things were done at FastMail if you like - we had a great system that which I really liked. For passkey we might want to find a way to simply call out to a limit aggregator instead of building something in.

Thanks! Yeah Id love to hear how you did it at FastMail. And also how we can simplify so it's enough for the magiclink+passkey PR to get merged.

PR #867, opened 2026-04-23
allow multiple route before functions

RensDimmendaal:
**Proposed Changes**
Allow multiple before functions in a route so you can do things like 

```py
@rt(before=(check_user, limit('20/d'))
def index(sess): ...
```

**Types of changes**

**Checklist**

**Additional Information**

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -20,7 +20,7 @@ from fastcore.meta import use_kwargs_dict,delegates
 from fastcore.style import S
 
 from types import UnionType, SimpleNamespace as ns, GenericAlias
-from typing import get_args, get_origin, Union, Mapping, List, Any
+from typing import get_args, get_origin, Union, Mapping, List, Any, Callable
 from datetime import datetime,date
 from dataclasses import dataclass
 from inspect import Parameter,get_annotations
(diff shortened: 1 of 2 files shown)


fasthtml/core.py
```diff
@@ -664,7 +664,7 @@ def add_route(self:FastHTML, route):
 
 # %% ../nbs/api/00_core.ipynb #26b147ba
 @patch
-def _endp(self:FastHTML, f, body_wrap, before=None):
+def _endp(self:FastHTML, f, body_wrap, before:Optional[Callable|tuple]=None):
```
jph00, line 667: I'm not a big fan of adding this complex typing tbh - not sure it's much of a win in dynamic environments, and adds clutter that (for me at least) makes it harder to see what's happening.

fasthtml/core.py
```diff
@@ -751,7 +752,7 @@ def _add_route(self:FastHTML, func, path, methods, name, include_in_schema, body
 
 # %% ../nbs/api/00_core.ipynb #f5cb2c2b
 @patch
-def route(self:FastHTML, path:str=None, methods=None, name=None, include_in_schema=True, body_wrap=None, host=None, before=None):
+def route(self:FastHTML, path:str=None, methods=None, name=None, include_in_schema=True, body_wrap=None, host=None, before:Optional[Callable|tuple]=None):
```
jph00, line 755: ...Ugh and it ends up cluttering every function it's threaded thru! :O

PR #871, opened 2026-05-02
Use @threaded in JupyUvi

kafkasl:
Moved `JupyUvi` to use `threaded`.

Decisions taken:

Decided to remove any thread reference. Using`daemon=True` fixes the original problem: if `.stop()` is never called, the leftover uvicorn thread won’t block process shutdown. 

Calling`join()` only helps when `.stop()` is called, by waiting for the server thread to actually finish. But we already call `wait_port_free()` so this extra `join(...)` is useful but not essential, and given that it requires extra thread code I decided to just remove it.

### DIFF
--- a/fasthtml/jupyter.py
+++ b/fasthtml/jupyter.py
@@ -13,7 +13,7 @@ from fastcore.utils import *
 from fastcore.meta import delegates
 from .common import *
 from .common import show as _show
-from fastcore.parallel import startthread
+from fastcore.parallel import threaded
 try: from IPython.display import HTML,Markdown,display
 except ImportError: pass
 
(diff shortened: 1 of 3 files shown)


comhar:
@kafkasl 

- can you use `startthread` (it now takes a [daemon param](https://github.com/AnswerDotAI/fastcore/commit/581eb7906ffdf1106898012817dc91cf1bb36d41))?
- can you bump the fastcore requirements to 1.12.45?
- can you include a usage example in the pr showing the problem?

kafkasl:
Addressed your comments @comhar , here you have the example.

In the first case with `daemon=False` when a test fails, the nbdev-test process never exits because it is waiting for the `server.stop()` call that will never arrive.

image

With `daemon=True` nbdev finishes despite the error.

image

jph00:
Thanks - this is much better! As close as possible to the original implementation, with just the smallest fix applied :)

PR #879, opened 2026-05-20
Improve `JupyUvi.stop` and async waiting

kafkasl:
The main goal of this PR is to fix some race conditions when calling `JupyUvi.stop`. I often get `address already busy` despite doing something like:

```
if 'server' in globals(): server.stop()
server.start()
```

It also addresses sync/async interference during shutdown by splitting cleanly those classes and start/stop methods.

## Issues

The main issues were:

1. Fixed host mismatch on stop checks: stop() now passes self.host into wait_port_free. We were waiting by default to `localhost` while uvicorn starts on `0.0.0.0`. Localhost interface can be free for biding "bindable" to a given port, while `0.0.0.0` (which requires ALL interfaces free) is still busy.
2. Made wait_port_free fail fast and explicit instead of silently printing: it now raises TimeoutError after timeout. The rationale is that whenever the stop wait fails, it is silently ignored and errors surface later which is more annoying.
3. Tightened port probe semantics (bind + listen and SO_REUSEADDR) so stop/restart logic tests a server-like
     listenability condition, not just socket creation.

All of this is explored and demonstrated with examples in this dialog https://share.solveit.pub/d/346001068041d03a6fd2c0b328473fa9

## Sync vs Async interfaces

In addition, I personally was a bit confused by the sync / async options. So I refactored. I am happy to change this as it introduces some breaking changes in the interface. 

The motivation was: 
```
  - JupyUvi = “I want fire-and-forget server startup in notebooks or scripts” (threaded, synchronous API).
  - JupyUviAsync = “I’m already inside an event loop” (async start/stop, no extra thread).
```

Mixing the two lead to another subtle bug:

`nb_serve_async`  runs in the notebook loop, but it used a blocking sync stop path (`time.sleep` vs `asyncio.sleep`) which prevented the shutdown coroutine work from progressing thus blocking uvicorn's shutdown work.

My solution was to split sync / async classes, and make the async.stop use `asyncio.sleep`.

## Notes

There's some extra output changes in the notebook due to the recent changes in SolveIT output rendering, I assume that is fine, but happy to surgically remove them if it is a bother


nbs/api/06_jupyter.ipynb
```diff
@@ -832,30 +778,11 @@
...
-     },
-     "execution_count": null,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
     "# Run the notebook locally to see the HTMX iframe in action\n",
-    "HTMX()"
```
kafkasl, line 858: I have commented this because it is designed to be run locally only, and completely blocks solveit. Happy to revert if you disagree

PR #880, opened 2026-05-23
Treat an empty application/json body as {} in parse_form

linguistic76:
## Problem

`parse_form` runs for every route (via `_wrap_req`) before handlers, to build the params
dict. It already guards an empty `multipart/form-data` body (returns `FormData()`), but the
`application/json` branch calls `await req.json()` directly. Starlette's `request.json()` does
`json.loads(b"")` on an empty body → `JSONDecodeError`, which is uncaught → **500**.

So a bodyless `POST` sent with `Content-Type: application/json` — common from HTTP clients/SDKs
that set that header by default — 500s before reaching the route, even for endpoints that take
no body.

### Reproduction

```python
from fasthtml.common import FastHTML
from starlette.testclient import TestClient

app = FastHTML()
@app.post("/x")
async def x(request): return "ok"

TestClient(app, raise_server_exceptions=False).post(
    "/x", headers={"content-type": "application/json"}
)  # 500
```

```
File ".../fasthtml/core.py", line ~165, in parse_form
    return await req.json() if ctype == 'application/json' else await req.form()
File ".../starlette/requests.py", line 259, in json
    self._json = json.loads(body)
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

(Starlette intentionally won't tolerate this in `request.json()` — Kludex/starlette#788 — so the
guard belongs here, alongside the existing empty-multipart one.)

## Fix

Mirror the empty-multipart guard for json: an empty json body yields `{}` (reusing the cached
body), non-empty parses as before. Other content types are unchanged.

```python
    body = await req.body()  # Cache body for non-multipart request types
    if ctype == 'application/json': return await req.json() if body else {}
    return await req.form()
```

Adds a test in `nbs/api/00_core.ipynb` asserting an empty `application/json` POST returns 200
with `{}` (instead of 500), and that a non-empty json body still parses to its dict (the
existing path is preserved). `nbdev_test --file_glob 00_core.ipynb` passes.


jph00:
Thanks!

PR #882, opened 2026-05-27
Update Starlette min version for BadHost CVE

ncoop57:


jph00:
Thanks - looks like we overlapped on fixing it! :)

PR #890, opened 2026-06-23
Fix Route handlers convert falsy values (0, False) to empty strings #572

### LINKED ISSUE #572: [BUG] Route handlers convert falsy values (0, False) to empty strings
**Describe the bug**
FastHTML route handlers convert falsy values to empty strings when rendering responses. Specifically:
- The integer `0` is converted to `''` (empty string) instead of `'0'`
- The boolean `False` is converted to `''` instead of `'False'`

This behavior is inconsistent with how other primitive values are handled, where integers like `1` correctly convert to `'1'` and `True` correctly converts to `'True'`.

```python
from fasthtml.core import FastHTML, Client

app, cli, rt = (app := FastHTML()), Client(app), app.route

@rt('/zero')
def get(): return 0

@rt('/one')
def get(): return 1

@rt('/true')
def get(): return True

@rt('/false')
def get(): return False

assert cli.get('/zero').text == '0'  # Fails: returns ''
assert cli.get('/one').text == '1'   # Passes
assert cli.get('/true').text == 'True'  # Passes
assert cli.get('/false').text == 'False'  # Fails: returns ''
```

**Expected behavior**
- Route handlers returning `0` should render as `'0'` in the response
- Route handlers returning `False` should render as `'False'` in the response

This would be consistent with how other primitive values are handled, where:
- `1` renders as `'1'`
- `True` renders as `'True'`

**Environment Information**
- fasthtml version: 0.9.0
- fastcore version: 1.7.19
- fastlite version: 0.0.13

**Confirmation**
- [X] I have included the versions of fastlite, fastcore, and fasthtml

**Additional context**
The current behavior makes it impossible to distinguish between an empty response and a response containing `0` or `False`.

Similar but unrelated bug #567, this one caused by line [188](https://github.com/fastai/fastcore/blob/a11d284feeb05e4eaa5deaab8aba38384665f094/fastcore/xml.py#L188C71-L188C97) `...v not in (False, None, '')...` in [to_xml](https://github.com/fastai/fastcore/blob/a11d284feeb05e4eaa5deaab8aba38384665f094/fastcore/xml.py#L168).

This happens because the initial check `if not resp: resp=()` in [`_resp()`](https://github.com/AnswerDotAI/fasthtml/blob/e9451caeb0560d275dfe5c7cace5feffa001aaf8/fasthtml/core.py#L406) treats these valid values as falsy conditions, preventing them from reaching the string conversion at the end of the function.

I don't know if this is intended, but i can imagine cases where `0` can be a valid response. At least, it think it should be deocumented so we can intercept this cases and return strings.

The above examples are contrived, but consider this simple component with a readout that never shows the `0` (running in a notebook):

```python
import dataclasses as DC
from fasthtml.components import Label, Input, Text

_n = '\n'

@DC.dataclass
class IntSlider:
    value:int=-1; min:int=-1; max:int=1; step:int=1
    def __ft__(self):
        return Div()(
            Label(_for='value')('Value'), _n,
            Input(type='range', name='value', **DC.asdict(self),
                hx_post=f"/value", hx_trigger='input changed', hx_target='next text', hx_swap='textContent'),
            Text(style='inline')(self.value), _n
        )

rp = IntSlider()

@rt('/value')
def changed(value:int):
    rp.value = value
    return value

show(rp)
```

**Screenshots**
Screenshot 2024-11-10 at 3 26 07 PM
Screenshot 2024-11-10 at 3 26 13 PM
Screenshot 2024-11-10 at 3 26 19 PM

civvic:
### Problem
Route handlers returning `0` or `False` rendered as an empty string instead of `'0'` / `'False'`. Values like `1` and `True` worked correctly, making the behaviour inconsistent.

### Cause
In `_resp` (`fasthtml/core.py`), the guard `if not resp: resp=''` is true for *all* falsy values. It's meant to catch "no response", but `0` and `False` are also falsy, so they were silently replaced with `''` before reaching the `str(resp)` conversion lower down.

### Fix
Narrow the check to only the value that actually means "no response":

```python
-    if not resp: resp=''
+    if resp is None: resp=''
```

Now `0` and `False` fall through to `resp = str(resp)` and render as `'0'` and `'False'`. `None` still renders empty, and empty containers (`()`, `[]`) are unaffected since they were never caught by this line — they flow through the FT-rendering path as before.

### Testing
```python
@rt('/zero')
def get(): return 0

assert cli.get('/zero').text == '0'
assert cli.get('/false').text == 'False'
assert cli.get('/one').text == '1'
assert cli.get('/true').text == 'True'
```

Closes #572.

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -458,7 +458,7 @@ def _is_ft_resp(resp):
 # %% ../nbs/api/00_core.ipynb #968d9245
 def _resp(req, resp, cls=empty, status_code=200):
     "Create appropriate HTTP response from request and response data"
-    if not resp: resp=''
+    if resp is None: resp=''
     if hasattr(resp, '__response__'): resp = resp.__response__(req)
     if not (isinstance(cls, type) and issubclass(cls, Response)): cls=empty
     if isinstance(resp, FileResponse) and not os.path.exists(resp.path): raise HTTPException(404, resp.path)
--- a/nbs/api/00_core.ipynb
+++ b/nbs/api/00_core.ipynb
@@ -1505,7 +1505,7 @@
     "#| export\n",
     "def _resp(req, resp, cls=empty, status_code=200):\n",
     "    \"Create appropriate HTTP response from request and response data\"\n",
-    "    if not resp: resp=''\n",
+    "    if resp is None: resp=''\n",
     "    if hasattr(resp, '__response__'): resp = resp.__response__(req)\n",
     "    if not (isinstance(cls, type) and issubclass(cls, Response)): cls=empty\n",
     "    if isinstance(resp, FileResponse) and not os.path.exists(resp.path): raise HTTPException(404, resp.path)\n",


jph00:
Thanks :)

PR #896, opened 2026-07-01
convert CamelCase tags to kebab-case

jackhogan:
handle CamelCase tag imports from `fasthtml.components` similarly to from `fastcore.xml`


jph00:
Maybe consider a followup PR which refactors the dupe code into fastcore so we don't end up with a similar issue next time we make a change like this? (This is my bad in the first place for reimplementing this!)

jackhogan:
Thanks @jph00, I refactored the duplicated code in fastcore so the implementation can be shared. See 
- https://github.com/AnswerDotAI/fastcore/pull/838
- https://github.com/AnswerDotAI/fasthtml/pull/899

PR #902, opened 2026-07-10
Add fast __deepcopy__ for FT to avoid slow generic object fallback

AmiraliOmidvar:
---
**Related Issue**
N/A (performance optimization)

**Proposed Changes**
Add a fast [__deepcopy__](cci:1://file:///home/amirali/work/mine/fasthtml/fasthtml/core.pyi:24:0-27:7) implementation for `FT` (Fast Tag) to avoid the slow generic object fallback. The implementation directly copies the five core fields (`tag`, `children`, `attrs`, `void_`, `listeners_`) while delegating unknown attr/child value types to `deepcopy` itself, so user-defined [__deepcopy__](cci:1://file:///home/amirali/work/mine/fasthtml/fasthtml/core.pyi:24:0-27:7) methods are still respected. This improves performance when copying FT components.

**Types of changes**

**Checklist**

**Additional Information**
Tests added in `nbs/api/00_core.ipynb` cover:
- Basic deepcopy with identity verification
- Nested structure mutation isolation
- Void element preservation
- Custom [__deepcopy__](cci:1://file:///home/amirali/work/mine/fasthtml/fasthtml/core.pyi:24:0-27:7) delegation
- `fast_app()` header/footer deepcopy
---

### DIFF
--- a/fasthtml/core.py
+++ b/fasthtml/core.py
@@ -40,6 +40,19 @@ from email.utils import format_datetime
 
 from .starlette import *
 
+# %% ../nbs/api/00_core.ipynb #2b18bf01
+@patch
+def __deepcopy__(self:FT, memo):
+    "Fast `deepcopy` for `FT`, avoiding the slow generic object fallback. Unknown attr/child value types are delegated to `deepcopy` itself, so a user's own `__deepcopy__` is still respected"
+    y = self.__class__.__new__(self.__class__)
+    memo[id(self)] = y
+    y.tag = self.tag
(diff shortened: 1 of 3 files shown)


jph00:
Thanks! We benchmarked this on fast_app's default headers: it saves ~19us/request, which doesn't show up against the cost of a real request. Since the fast path also silently drops extra instance state on FT subclasses (the generic fallback copies __dict__ faithfully), we're going to pass. If header copying ever shows in a real profile, we'd remove the per-request deepcopy instead.

PR #909, opened 2026-07-20
Fix duplicate HTMX component parameters

### LINKED ISSUE #529: [BUG] Duplicated attributes in components.pyi
The attributes `hx_replace_url` and `hx_indicator` are displayed twice in each component signature

https://github.com/AnswerDotAI/fasthtml/blob/f0bde8c93854e01e364fbf7c3a062c0cd639664a/fasthtml/components.pyi#L75

Affected version: In current main branch

### LINKED ISSUE #908: [BUG] Duplicate parameter "hx_indicator" in function definition in fasthtml/components.pyi
**Describe the bug**
in  fasthtml/components.pyi there are 115 function definitions where hx_indicator is duplicated, that makes the file not correctly readable by `mypy` that throws  `error: Duplicate parameter "hx_indicator" in function definition`

On https://mypy-play.net/?gist=e99d5734856f8ac11110547eea074de5 I have copied the file in the mypy playground

erikgaas:
## Summary

- Remove duplicate `indicator` and `replace_url` entries from `hx_attrs`.
- Add a notebook test requiring the exported attribute list to contain unique names.
- Regenerate `components.pyi` without duplicate parameters.

## Why

`tools/mk_pyi.py` expands `hx_attrs` directly into every component signature. Duplicate list entries therefore generate duplicate `hx_indicator` and `hx_replace_url` parameters, which makes the stub invalid for mypy.

Runtime component signatures already deduplicate these names, so this change does not alter rendered HTML or supported arguments.

Fixes #908.
Fixes #529.

## Checks

- `nbdev-test --path nbs/api/01_components.ipynb`
- Parsed `fasthtml/components.pyi` with `ast.parse`
- Confirmed no generated signature repeats `hx_indicator` or `hx_replace_url`
- Confirmed a fresh interpreter still renders both attributes correctly


jph00:
Thanks! BTW @erikgaas I feel like your test is either over-enthusiastic (does this really need a test) or under-enthusiastic (if hx is being tested, why not all of them). Just testing the one thing you changed doesn't seem like the right answer AFAICT! :D

